PR #4, opened 2019-11-14
feat: render view context content

shauke:


src/app/core/services/cms/cms.service.ts
```diff
@@ -50,4 +52,32 @@ export class CMSService {
...
+  getViewContextContent(
+    viewContextId: string,
+    callParameters: CallParameters
+  ): Observable<{ entrypoint: ContentPageletEntryPoint; pagelets: ContentPagelet[] }> {
+    if (!viewContextId) {
+      return throwError('getViewContextContent() called without a viewContextId');
+    }
+
+    let params = new HttpParams();
+    Object.keys(callParameters).forEach(item => (params = params.set(item, callParameters[item])));
```
dhhyi, line 71: This could become a beautiful one-liner using reduce: `const params = Object.entries(callParameters).reduce((p, [key, value]) => p.set(key, value), new HttpParams());`

src/app/core/store/content/viewcontexts/viewcontexts.effects.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        provideMockActions(() => actions$),
+        { provide: CMSService, useFactory: () => instance(cmsServiceMock) },
+      ],
+    });
+
+    effects = TestBed.inject(ViewcontextsEffects);
+  });
+
+  describe('loadViewContextEntrypoint$', () => {
+    it('should not dispatch actions when encountering loadViewcontexts', () => {
```
dhhyi, line 34: "should not dispatch"? 🤔

src/app/core/store/content/viewcontexts/viewcontexts.effects.ts
```diff
@@ -0,0 +1,36 @@
...
+export class ViewcontextsEffects {
+  constructor(private actions$: Actions, private cmsService: CMSService) {}
+
+  loadViewContextEntrypoint$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(loadViewContextEntrypoint),
+      mapToPayload(),
+      switchMap(({ viewContextId, callParameters }) =>
+        this.cmsService.getViewContextContent(viewContextId, callParameters).pipe(
+          log('view context success'),
```
dhhyi: dead code?

src/app/core/store/content/viewcontexts/viewcontexts.effects.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable()
+export class ViewcontextsEffects {
+  constructor(private actions$: Actions, private cmsService: CMSService) {}
+
+  loadViewContextEntrypoint$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(loadViewContextEntrypoint),
+      mapToPayload(),
+      switchMap(({ viewContextId, callParameters }) =>
```
dhhyi, line 22: please use `concatMap`. `switchMap` would discard all actions but the last.

src/app/core/store/content/viewcontexts/viewcontexts.effects.ts
```diff
@@ -0,0 +1,36 @@
...
+            loadViewContextEntrypointSuccess({ entrypoint, pagelets, viewContextId, callParameters })
+          ),
+          mapErrorToAction(loadViewContextEntrypointFail)
+        )
+      )
+    )
+  );
+}
+
+//           map(([newRoles]) => setUserRolesSuccess({ login, roles: newRoles })),
```
dhhyi: dead code?

src/app/core/store/content/viewcontexts/viewcontexts.reducer.ts
```diff
@@ -0,0 +1,55 @@
...
+declare type ContentPageletEntryPointWithContext = ContentPageletEntryPoint & {
+  viewContextId: string;
+  callParameters: CallParameters;
+};
+
+export function serializeContextSpecificViewContextId(viewContextId: string, callParameters: CallParameters) {
+  return (
+    viewContextId +
+    Object.keys(callParameters)
+      .map(key => `@${key}-${callParameters[key]}`)
```
dhhyi, line 23: As you need both key and value it might be better to use `Object.entires`.
On a different note: it might be also good to sort the keys, so view contexts are unique independent of call parameters order.
\+ add test for this 😉

src/app/core/store/content/viewcontexts/viewcontexts.reducer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const viewcontextsAdapter = createEntityAdapter<ContentPageletEntryPointWithContext>({
+  selectId: viewcontext => serializeContextSpecificViewContextId(viewcontext.viewContextId, viewcontext.callParameters),
+});
+
+export interface ViewcontextsState extends EntityState<ContentPageletEntryPointWithContext> {
+  loading: boolean;
+}
+
+export const initialState: ViewcontextsState = viewcontextsAdapter.getInitialState({
```
dhhyi, line 36: export is not needed, right?

src/app/pages/category/category-categories/category-categories.component.html
```diff
@@ -12,6 +12,14 @@
 
     <div class="col-md-9">
       <ish-breadcrumb></ish-breadcrumb>
+
+      <!-- ViewContext usage example
+      <ish-content-viewcontext
+        viewContextId="vc_category_detail"
+        [callParameters]="{ Category: category.categoryRef }"
+      ></ish-content-viewcontext>
+      -->
```
dhhyi, line 21: dead code?

src/app/pages/category/category-products/category-products.component.html
```diff
@@ -17,6 +17,14 @@
...
     <div class="col-md-9">
       <ish-breadcrumb></ish-breadcrumb>
+
+      <!-- ViewContext usage example
+      <ish-content-viewcontext
+        viewContextId="vc_category_detail"
+        [callParameters]="{ Category: category.categoryRef }"
+      ></ish-content-viewcontext>
+      -->
+
```
dhhyi, line 27: dead code?

src/app/pages/product/product-page.component.html
```diff
@@ -5,6 +5,13 @@
     <ng-container *ngIf="product$ | async as product">
       <ish-breadcrumb></ish-breadcrumb>
 
+      <!-- ViewContext usage example
+      <ish-content-viewcontext
+        viewContextId="vc_product_detail"
+        [callParameters]="{ Product: product.sku, Category: (category$ | async)?.categoryRef }"
+      ></ish-content-viewcontext>
+      -->
```
dhhyi, line 13: Do we not plan to just set those active for the demo deployment? I'd rather not have dead code...

src/app/core/store/content/viewcontexts/viewcontexts.reducer.ts
```diff
@@ -0,0 +1,55 @@
...
+    loading: false,
+  })),
+  on(loadViewContextEntrypointSuccess, (state: ViewcontextsState, action) => {
+    const { entrypoint, viewContextId, callParameters } = action.payload;
+
+    return {
+      ...viewcontextsAdapter.upsertOne({ ...entrypoint, viewContextId, callParameters }, state),
+      loading: false,
+    };
+  })
```
dhhyi, line 54: please use `unsetLoadingAndErrorOn` and use two handlers for each action. That way entity handlers are separated from loading handlers. This method should also work for states that don't have error properties.

See src/app/extensions/quoting/store/quoting/quoting.reducer.ts

src/app/core/store/content/viewcontexts/viewcontexts.selectors.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { serializeContextSpecificViewContextId, viewcontextsAdapter } from './viewcontexts.reducer';
+
+const getViewcontextsState = createSelector(getContentState, state => state.viewcontexts);
+
+export const getViewcontextsLoading = createSelector(getViewcontextsState, state => state.loading);
+
+export const { selectEntities: getViewcontextEntities } = viewcontextsAdapter.getSelectors(getViewcontextsState);
+
+const getViewcontextInternal = (viewContextId: string, callParameters: CallParameters) =>
```
dhhyi, line 15: Maybe add comment, that this selector is output-memoized. Alternatively rename it to `getViewcontextMemoized` (prefer the latter) 😉

src/app/pages/product/product-page.component.html
dhhyi: Also, please subscribe only once to `category$` in the template, when activating this.

dhhyi (changes requested):

src/app/pages/product/product-page.component.html
shauke: Any good idea how to do that? I did not find a way besides using `*ngIf` which I did not want to use since the `category$` is more or less an optional value?

dhhyi: Ah, yes, this is a problem. Maybe it is finally time to properly import https://ngrx.io/guide/component/let

shauke: We cannot activate the ViewContexts as long as they are not system managed in ICM meaning there are probably no fitting ViewContexts for the given 'viewContextId'. So I will remove the dead code for now an just leave the documentation example in `content-viewcontext.component.ts`.

shauke: This "solves" the category problem too for now. ;)

src/app/core/store/content/viewcontexts/viewcontexts.effects.spec.ts
shauke: left over from generated test

src/app/core/store/content/viewcontexts/viewcontexts.reducer.ts
```diff
@@ -0,0 +1,42 @@
...
+declare type ContentPageletEntryPointWithContext = ContentPageletEntryPoint & {
+  viewContextId: string;
+  callParameters: CallParameters;
+};
+
+export function serializeContextSpecificViewContextId(viewContextId: string, callParameters: CallParameters) {
+  const serializedParams = callParameters
+    ? Object.entries(callParameters)
+        .map(([key, value]) => `@@${key}-${value}`)
+        .sort()
```
dhhyi, line 18: technically, sorting should be done before serializing....

dhhyi (dismissed):

PR #8, opened 2019-11-19
feat: translations with links and callback (ISREST-915)

shauke:


src/app/core/directives/server-html.directive.ts
```diff
@@ -87,14 +91,21 @@ export class ServerHtmlDirective implements AfterContentInit, AfterViewInit, OnD
...
 
         // apply default link handling for empty href, external links & target _blank
-        if (!href || href.startsWith('http') || el.getAttribute('target') === '_blank') {
+        if (!cb && (!href || href.startsWith('http') || el.getAttribute('target') === '_blank')) {
           return;
         }
 
-        // otherwise handle as routerLink
-        this.router.navigateByUrl(href);
+        if (cb) {
```
fmalcher: Also check for existence of the callback before blindly calling it:

```
if (cb && this.callbacks && typeof this.callbacks[cb] === 'function')
```

src/app/core/directives/server-html.directive.ts
```diff
@@ -21,6 +21,10 @@ import { LinkParser } from 'ish-core/utils/link-parser';
   selector: '[ishServerHtml]',
 })
 export class ServerHtmlDirective implements AfterContentInit, AfterViewInit, OnDestroy, OnChanges {
+  @Input() callbacks?: {
```
fmalcher: I think, the optional mark `?` is not necesssary here since this property can remain `undefined`.

src/app/pages/registration/components/registration-form/registration-form.component.html
```diff
@@ -59,25 +59,27 @@ <h2>
...
-                <span>{{ 'registration.tac_privacy_policy.begin.text' | translate }} </span>
-                <ish-modal-dialog-link
-                  linkText="registration.tac.link"
-                  [options]="{ titleText: 'checkout.termsandconditions.details.title' | translate, size: 'lg' }"
-                >
+                <span
+                  [ishServerHtml]="'registration.tac_privacy_policy.label' | translate"
+                  [callbacks]="{
+                    callbackTAC: showModalDialog(modalDialogTAC),
+                    callbackPP: showModalDialog(modalDialogPP)
```
dhhyi: this would also work:
```
                  [callbacks]="{
                    callbackTAC: modalDialogTAC.show,
                    callbackPP: modalDialogPP.show
                  }"

```

That way the method in the component file would not be needed.

dhhyi: I seem to have mis-tested. :sweat_smile:

dhhyi:
@shauke Please merge when ready :wink:

PR #11, opened 2019-11-21
Spike: SEO-friendly URLs (ISREST-925,#33)

fmalcher:
This PR prepares custom URL formats for products and categories.
It consists of two parts:

1) Provide custom matchers for product and category routes that define the same URL format like before. Thus, this is non-breaking and could be merged. ⚠️ **Important for review: Look at the new structure!**

2) Define a second implementation for the matchers (and URL generators) to provide a different URL pattern. Caution: No tests added here because this is just a concept proof. Once we agreed on a URL format, we should also add tests for the matcher.

```
http://localhost:4200/Cameras-Camcorders.832-c
http://localhost:4200/Cameras-Camcorders.832-c/Google-Home-p-201807171.html
http://localhost:4200/Google-Home-p-201807171.html
```

### TODO:

* [ ] Agree on a custom URL format
* [ ] Implement it
* [ ] Add tests for the matcher and generator function
* [ ] Find out if there are places where URLs are still hard coded
* [ ] Add documentation on how our matcher format works
* [ ] Squash branch since the commit history is not clean

# Notes for documentation

**TODO: Find a better place for this!**

- use a custom matcher when the URL can't be matched by the simple standard path format
- set the matcher in the route via `matcher` property
- for each custom route format create an object of type `CustomRoute`.
- `CustomRoute` contains the matcher, a function for generating route URLs and all possible matched formats
  - matcher: used in the route
  - URL generator: used for creating URLs based on entities, used in `RouterLink` and `router.navigateByUrl()`
  - format list: used in effects to match the route

## Matcher function
- of type `UrlMatcher` by Angular
- Arguments:
  - array of UrlSegments (parts of the URL separated by slashes)
  - the route
- matcher must use its arguments to decide whether the route will match or not
- matcher must set `route.data` to empty object of no route data is defined
- on no match: return `null`/`undefined`
- on match:
  - set `route.data.format` to the actually matched route format, e.g. `product/:sku/**`. Also add this format to the list of all possible formats in the `CustomRoute` object
  - return object with properties:
    - `posParams`: all extracted params from the route, like `sku` or `categoryUniqueId`. Each as instance of `UrlSegment`
    - `consumed`: the part of the path that was covered by the matcher. This is necessary for when a sub-module contains other routes. Usually not interesting in our scenarios, thus returns the whole array of UrlSegments (`url`) in most cases


fmalcher:
TODO:

[ ] Agree on a custom URL format
[ ] Implement it
[ ] Add tests for the matcher and generator function

fmalcher:
@shauke pointed out that `ofRoute()` also sues the route format. Things got a bit tricky here: The `ofRoute()` mechanics use the `path` property from the route to determine the full path used for `ofRoute()`.

I needed to make the route (!) contain the actually matched path format. Since the matcher decides this on its own we don't know this before the matcher has run.
The only place where we could reliably store such information is the route data. Thus, the matcher now writes its decision to the route data, and the `ofRoute()` operator uses this data to get the full path.

I also introduced a custom route format that contains matcher, URL generator and possible formats. This list of formats can be used in effects then to match with `ofRoute()`.

Possible TODO:
Generalize this a bit more! The route formats are duplicated now: once in the matcher, once in the list of possible formats.

dhhyi:
waiting for possible interaction with mapping library (https://github.com/intershop/intershop-pwa/projects/2#card-31666961)

dhhyi:
Ideas integrated into #110

PR #17, opened 2019-11-26
fix: production mode support for concardis credit card

SGrueber:


src/app/pages/checkout-payment/components/payment-concardis-creditcard/payment-concardis-creditcard.component.ts
```diff
@@ -186,33 +191,40 @@ export class PaymentConcardisCreditcardComponent implements OnInit, OnChanges, O
...
-
-      this.errorMessage.cvc = error.message.properties.find(prop => prop.key === 'verification');
-      if (this.errorMessage.cvc && this.errorMessage.cvc.code) {
-        this.errorMessage.cvc.messageKey = this.getErrorMessage(
-          this.errorMessage.cvc.code,
-          'cvc',
-          this.errorMessage.cvc.message
-        );
-      }
+      if (typeof error.message !== 'string' && error.message && error.message.properties) {
```
dhhyi: I think when `typeof error.message !== 'string'` is true, then there has to be an object, so the following `error.message` is true at this point

src/app/core/services/script-loader/script-loader.service.ts
```diff
@@ -12,16 +14,21 @@ export class ScriptLoaderService {
...
+      src: 'https://pp.payengine.de/bridge/1.0/payengine.min.js',
+      loaded: false,
+    },
   ];
 
   /**
    * load a script, if it has not already been loaded
    */
-  load(name: string): Observable<ScriptType> {
+  load(name: ScriptName): Observable<ScriptType> {
```
dhhyi: this makes the `ScriptLoaderService` really special and makes further abstraction harder. I'd favor that the payment-concardis-creditcard.component knows which script to load and the service is more abstract

src/app/pages/checkout-payment/components/payment-concardis-creditcard/payment-concardis-creditcard.component.ts
```diff
@@ -117,9 +117,14 @@ export class PaymentConcardisCreditcardComponent implements OnInit, OnChanges, O
         return;
       }
 
+      const environment =
```
dhhyi: the name `environment` causes a lint warning

src/app/core/services/script-loader/script-loader.service.ts
```diff
@@ -1,29 +1,29 @@
 import { Injectable } from '@angular/core';
 import { Observable, Observer } from 'rxjs';
 
+export type ScriptName = 'concardis-payengine-sandbox' | 'concardis-payengine-production';
```
dhhyi: now this is unused :sweat_smile:

src/app/core/services/script-loader/script-loader.service.ts
```diff
@@ -1,29 +1,29 @@
...
-    },
-  ];
+  private scripts: ScriptType[] = [];
 
   /**
    * load a script, if it has not already been loaded
+   * @param url  script url, e.g. https://pptest.payengine.de/bridge/1.0/payengine.min.js
    */
-  load(name: string): Observable<ScriptType> {
+  load(url: string): Observable<ScriptType> {
```
dhhyi: Great. I like it :+1:

SGrueber: Thx :)

PR #25, opened 2019-11-28
refactor: use ngrx/entity in quoting feature

fmalcher:
This is the starting point for the quoting refactoring.

First steps done:
* [x] Use ngrx/entity for quotes and quote requests
* [x] Refactor selectors to be easier to reason about

### DIFF
--- a/src/app/extensions/quoting/store/quote-request/quote-request.selectors.ts
+++ b/src/app/extensions/quoting/store/quote-request/quote-request.selectors.ts
@@ -4,79 +4,69 @@ import { isEqual } from 'lodash-es';
 import { getProductEntities } from 'ish-core/store/shopping/products';
 
 import { QuoteRequestHelper } from '../../models/quote-request/quote-request.helper';
+import { QuoteRequestData } from '../../models/quote-request/quote-request.interface';
 import { getQuotingState } from '../quoting-store';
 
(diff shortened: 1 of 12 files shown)


src/app/extensions/quoting/store/quote-request/quote-request.selectors.ts
```diff
@@ -4,79 +4,69 @@ import { isEqual } from 'lodash-es';
...
+    items.map(item => ({
+      ...item,
+      product: item.productSKU ? products[item.productSKU] : undefined,
+    }))
+);
+
+export const getActiveQuoteRequest = createSelector(
+  getCurrentQuoteRequests,
+  quoteRequests => {
+    const quoteRequest = quoteRequests.find(item => item.editable);
```
fmalcher: I would not move the active/editable info to the state (e.g. as a `activeQuoteRequestId`) because this would be derived state. Is it possible that more than 1 quote request is editable?

If there is only 1 at a time, we could move this: Remove the `editable` flag from the QR and introduce a state prop `activeQuoteRequestId`.
Not sure...

Sebastian-Haehnlein: more than one quote requests are possibly editable. i will change this line to get the last hit.

PR #42, opened 2019-12-09
fix: show/hide logic for addToCard, quantity, addToQuoteRequest

Sebastian-Haehnlein:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring 
- [ ] Build-related changes
- [ ] CI-related changes
- [ ] Documentation content changes
- [ ] application / infrastructure changes
- [ ] Other... Please describe:

Issue Number: ISREST-951

## What Is the New Behavior?

hide addToCard, quantity, addToQuoteRequest in some cases

added criteria product.salePrice

refactored (cleanup, dont repeat yourself...)

- [ ] Yes
- [x] No

### DIFF
--- a/src/app/core/models/product/product.helper.ts
+++ b/src/app/core/models/product/product.helper.ts
@@ -181,4 +181,13 @@ export class ProductHelper {
     const attributes = product.attributes.filter(att => !common.includes(att.name));
     return { ...product, attributes };
   }
+
+  static showAddToCart(product: Product): boolean {
+    return !!(
+      product.inStock &&
+      product.availability &&
+      !ProductHelper.isMasterProduct(product) &&
+      (product.salePrice || ProductHelper.isRetailSet(product))
+    );
+  }
 }
--- a/src/app/core/models/product/product.helper.ts
+++ b/src/app/core/models/product/product.helper.ts
@@ -181,13 +181,4 @@ export class ProductHelper {
     const attributes = product.attributes.filter(att => !common.includes(att.name));
     return { ...product, attributes };
   }
-
-  static showAddToCart(product: Product): boolean {
-    return !!(
-      product.inStock &&
-      product.availability &&
-      !ProductHelper.isMasterProduct(product) &&
-      (product.salePrice || ProductHelper.isRetailSet(product))
-    );
-  }
 }
--- a/src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
+++ b/src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
@@ -3,7 +3,7 @@ import { BehaviorSubject, Observable, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 
 import { CheckoutFacade } from 'ish-core/facades/checkout.facade';
-import { Product } from 'ish-core/models/product/product.model';
+import { Product, ProductHelper } from 'ish-core/models/product/product.model';
 import { whenFalsy } from 'ish-core/utils/operators';
 
 /**
@@ -51,6 +51,8 @@ export class ProductAddToBasketComponent implements OnInit, OnDestroy {
    * button was clicked event
    */
   @Output() productToBasket = new EventEmitter<void>();
+
+  showAddToCart = ProductHelper.showAddToCart;
 
   constructor(private checkoutFacade: CheckoutFacade) {}
 
--- a/src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
+++ b/src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
@@ -52,8 +52,6 @@ export class ProductAddToBasketComponent implements OnInit, OnDestroy {
    */
   @Output() productToBasket = new EventEmitter<void>();
 
-  showAddToCart = ProductHelper.showAddToCart;
-
   constructor(private checkoutFacade: CheckoutFacade) {}
 
   // fires 'true' after add To Cart is clicked and basket is loading
@@ -84,5 +82,14 @@ export class ProductAddToBasketComponent implements OnInit, OnDestroy {
 
   ngOnDestroy() {
     this.destroy$.next();
+  }
+
+  hasAddToBasketButton() {
+    return !!(
+      this.product.inStock &&
+      this.product.availability &&
+      !ProductHelper.isMasterProduct(this.product) &&
+      (this.product.salePrice || ProductHelper.isRetailSet(this.product))
+    );
   }
 }
--- a/src/app/shared/components/product/product-row/product-row.component.html
+++ b/src/app/shared/components/product/product-row/product-row.component.html
@@ -52,14 +52,8 @@
           [product]="product"
           [showInformationalPrice]="true"
         ></ish-product-price>
-        <ish-product-inventory
-          *ngIf="is('displayInventory') && !isMasterProduct(product)"
-          [product]="product"
-        ></ish-product-inventory>
-        <ish-product-shipment
-          *ngIf="is('displayShipment') && !isMasterProduct(product)"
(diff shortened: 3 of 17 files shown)


src/app/core/models/product/product.helper.ts
```diff
@@ -151,4 +151,13 @@ export class ProductHelper {
       };
     }
   }
+
+  static showAddToCart(product: Product): boolean {
+    return !!(
+      product.inStock &&
+      product.availability &&
+      !ProductHelper.isMasterProduct(product) &&
+      (product.salePrice || ProductHelper.isVariationProduct(product) || ProductHelper.isRetailSet(product))
```
dhhyi: what is the logic of this line?

Show if it has a sale price -- ok
Show if it is a variation product or retail set even if it has no price? -- is that even possible?

src/app/shared/product/components/product-add-to-basket/product-add-to-basket.component.html
```diff
@@ -1,4 +1,4 @@
-<div *ngIf="product.inStock && product.availability" class="add-to-cart">
+<div *ngIf="showAddToCart(product)" class="add-to-cart">
```
dhhyi: we have a MR in our GitLab (!847), which I would like to see this dependant on

src/app/core/models/product/product.helper.ts
```diff
@@ -151,4 +151,13 @@ export class ProductHelper {
       };
     }
   }
+
+  static showAddToCart(product: Product): boolean {
```
dhhyi: The location of this method seems wrong. This is somehow mixing concerns.

src/app/core/models/product/product.helper.ts
Sebastian-Haehnlein: retailset has other attributes: summedUpSalePrice, minSalePrice
the condition for variation-products is not really necessary (removed)

dhhyi: okay :+1:

src/app/shared/components/product/product-tile/product-tile.component.html
```diff
@@ -6,24 +6,24 @@
     <ish-product-label [product]="product"></ish-product-label>
   </div>
 
-  <a *ngIf="configuration.displayName" class="product-title" [routerLink]="product | ishProductRoute: category">{{
+  <a *ngIf="is('displayName')" class="product-title" [routerLink]="product | ishProductRoute: category">{{
     product.name
   }}</a>
 
   <ish-product-rating *ishFeature="'rating'" [product]="product"></ish-product-rating>
```
dhhyi: rating display should also be configurable.

src/app/shared/components/product/product-tile/product-tile.component.html
```diff
@@ -34,20 +34,20 @@
```
dhhyi: this one above could become a switch for `displayVariationCount`

src/app/shared/components/product/product-row/product-row.component.html
```diff
@@ -11,31 +11,31 @@
   <div class="col-9 col-md-10">
     <div class="row">
       <div class="col-md-7 col-xs-12">
-        <a *ngIf="configuration.displayName" class="product-title" [routerLink]="product | ishProductRoute: category">{{
+        <a *ngIf="is('displayName')" class="product-title" [routerLink]="product | ishProductRoute: category">{{
           product.name
         }}</a>
 
         <ish-product-rating *ishFeature="'rating'" [product]="product"></ish-product-rating>
```
dhhyi: rating display should also be configurable

src/app/shared/components/product/product-row/product-row.component.html
```diff
@@ -46,21 +46,21 @@
...
       <div class="col-12 col-md-5 text-md-right">
         <ish-product-price
-          *ngIf="configuration.displayPrice"
+          *ngIf="is('displayPrice')"
           [product]="product"
           [showInformationalPrice]="true"
         ></ish-product-price>
         <ish-product-inventory
-          *ngIf="configuration.displayInventory && !isMasterProduct(product)"
+          *ngIf="is('displayInventory') && !isMasterProduct(product)"
```
dhhyi: `!isMasterProduct` should be part of the configuration function

src/app/shared/components/product/product-row/product-row.component.html
```diff
@@ -46,21 +46,21 @@
...
           [product]="product"
         ></ish-product-inventory>
         <ish-product-shipment
-          *ngIf="configuration.displayShipment && !isMasterProduct(product)"
+          *ngIf="is('displayShipment') && !isMasterProduct(product)"
```
dhhyi: `!isMasterProduct` should be part of the configuration function

src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
```diff
@@ -83,4 +83,13 @@ export class ProductAddToBasketComponent implements OnInit, OnDestroy {
   ngOnDestroy() {
     this.destroy$.next();
   }
+
+  hasAddToBasketButton() {
```
dhhyi: Why here? Isn't this already decided by parent components?

Sebastian-Haehnlein: not really, ProductAddToBasketComponent is used in six other components and just "tile" and "row" has this condition in the ProductItemContainerConfiguration

src/app/shared/components/product/product-tile/product-tile.component.html
Sebastian-Haehnlein: ?

src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
dhhyi: But then the logic is executed twice for product-rows and product-tiles, which is not what we want. Either we do it always from the outside or we do it from inside the component.

It seems that in ISML it is done in the component itself. But then there would be no need to have configuration from the outside.

So we either do it always from the outside (then we'll need a helper method after all) or we only let the add-to-basket component decide when to show itself.

opinions? @shauke

shauke:
Something is not right with the quantity labels in the product listing in row view.
quantity-label-error

Sebastian-Haehnlein:
> Something is not right with the quantity labels in the product listing in row view.
> quantity-label-error

fixed

src/app/shared/components/product/product-add-to-basket/product-add-to-basket.component.ts
shauke: The issue whether to show the Add To Cart button or not has two aspects.

It should be configurable at parent components like the product tile, where we might want to use the same product tile component at several places with different rendering styles showing the add to cart button at some places and not at others. So this needs to be for some parent components a display configuration (e.g. for the product tile and product row, but not for product detail or product compare). This is the case in ISML too (see ProductTile.isml).

On the other hand there needs to be logic at the Add To Cart component itself too, that determines whether the add to cart button can be show or not. This logic should be independent from the parents configuration to show the add to cart button. This logic is only based on product properties that are needed to show an add to cart button or not. This is the same in ISML too (see AddToCartInc.isml). And it should be independent from the parent so the Add To Cart component can be used in several places without the need to implement the same logic whether to display it or not in several parent components (compare listing, product tile, product row, product detail, ...).

dhhyi: So as an example: The check for availability (inStock, ...) should be made within the component and disabling the button for master products should be done from the outside?

dhhyi:
Replaced by #403

PR #51, opened 2019-12-11
fix: error messages quote-edit

shauke:

Confusing error message is displayed: "400: Bad Request (Adding specified item finished with an error.)"

Out of Stock or Offline Product

(Issue Number: ISREST-948)

## What Is the New Behavior?

* Error message should display: The quote cannot be added to the shopping cart. It contains invalid products.
* show Availability 
* remove "Add to Cart" button

IAD_offline_products

## Steps to Repeat
* Add product to quote request
* Submit quote request
* Submit quote via backoffice
* Set product to out of stock or offline
* View responded quote

[x] No

### DIFF
--- a/src/app/extensions/quoting/facades/quoting.facade.ts
+++ b/src/app/extensions/quoting/facades/quoting.facade.ts
@@ -13,6 +13,7 @@ import {
   DeleteQuote,
   LoadQuotes,
   RejectQuote,
+  ResetQuoteError,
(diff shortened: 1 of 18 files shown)


shauke:
In regards to the original issue, there is still a lot to do and even the current fix itself only changed the format of the error message a bit but not in the requested way. Or was this discussed with IAD?

PR #66, opened 2019-12-17
feat: seo support for robots.txt

Sebastian-Haehnlein:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] CI-related changes
- [ ] Documentation content changes
- [ ] application / infrastructure changes
- [ ] Other... Please describe:

Issue Number: 

- [ ] Yes
- [x] No

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -20,6 +20,8 @@ import 'zone.js/dist/zone-node';
 
 import * as express from 'express';
 import { join } from 'path';
+import * as robots from 'express-robots-txt';
+import * as fs from 'fs';
 
 const logging = !!process.env.LOGGING;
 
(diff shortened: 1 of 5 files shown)


server.ts
```diff
@@ -44,6 +46,36 @@ app.engine(
 app.set('view engine', 'html');
 app.set('views', join(DIST_FOLDER, 'browser'));
 
+// seo robots.txt
+app.use((req, res, next) => {
```
dhhyi: how about using it like it is documented:
```
if (fs.existsFileSync(...robots)) {
    app.use(robots(..file..))
} else {
    app.use(robots(..object..))
}

```

angular.json
```diff
@@ -29,7 +29,8 @@
             "assets": [
               "src/favicon.ico",
               "src/assets",
-              "src/manifest.webmanifest"
+              "src/manifest.webmanifest",
+              "src/robots.txt"
```
dhhyi: we only want to have robots available with ssr. No assets needed. Just put it in project root.

PR #72, opened 2019-12-18
fix: ngx-meta translate on load is not working

### LINKED ISSUE #69: [QUESTION] SEO locales
I'm currently using the "seo.title.home" in the app-routing-module.ts. 

It's always showing me the "en_US" translation as my title. The weird thing is that I expect to see another translated title here ("nl_NL") since I'm also seeing my other translations perfectly fine in my components while using the translate pipe.

Am I forgetting to set a default language somewhere? Or are the seo related things always looking at "en_US"?

I have defined my own language "nl_NL" inside all my environments as the first locale.

Thank you in advance.

Sebastian-Haehnlein:
## PR Type

- Bugfix

## Issue Number: 
Closes #69

### DIFF
--- a/src/app/core/store/locale/locale.effects.spec.ts
+++ b/src/app/core/store/locale/locale.effects.spec.ts
@@ -1,10 +1,10 @@
 import { TestBed } from '@angular/core/testing';
+import { ROOT_EFFECTS_INIT } from '@ngrx/effects';
 import { provideMockActions } from '@ngrx/effects/testing';
 import { Action, Store } from '@ngrx/store';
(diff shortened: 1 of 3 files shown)


dhhyi:
I like the approach, but this will just postpone the issue. Locales shall come from the ICM in the distant future. Let's take it agile as it is. :+1:

Sebastian-Haehnlein:
:santa:

dhhyi:
:christmas_tree:

dhhyi:
@nickyhaze this should do it, right?

nickyhaze:
It does! Thank you very much

PR #112, opened 2020-02-12
fix: quote dialog submit behaviour

Sebastian-Haehnlein:
## PR Type

 [x] Bugfix  

"Submit Quote Request" links to quote request details in my account area. The user loses his context.
(Issue Number: ISREST-941)

## What Is the New Behavior?

Click on "Submit Quote Request" - closes modal and shows "Thank you for your quote request" in a new modal. 
(Additional: Behaviour of "Create new QuoteRequest from QuoteRequest" in Dialog)

 [ ] Yes  
 [x] No

### DIFF
--- a/src/app/extensions/quoting/shared/product/product-add-to-quote-dialog/product-add-to-quote-dialog.component.ts
+++ b/src/app/extensions/quoting/shared/product/product-add-to-quote-dialog/product-add-to-quote-dialog.component.ts
@@ -2,10 +2,12 @@ import { ChangeDetectionStrategy, Component, OnDestroy, OnInit } from '@angular/
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap';
 import { Observable, Subject } from 'rxjs';
-import { distinctUntilKeyChanged, filter, take, takeUntil } from 'rxjs/operators';
+import { take, takeUntil } from 'rxjs/operators';
 
+import { AccountFacade } from 'ish-core/facades/account.facade';
 import { MessageFacade } from 'ish-core/facades/message.facade';
 import { LineItemUpdate } from 'ish-core/models/line-item-update/line-item-update.model';
+import { User } from 'ish-core/models/user/user.model';
 import { whenTruthy } from 'ish-core/utils/operators';
 
 import { QuotingFacade } from '../../../facades/quoting.facade';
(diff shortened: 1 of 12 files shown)


src/app/extensions/quoting/shared/product/product-add-to-quote-dialog/product-add-to-quote-dialog.component.ts
```diff
@@ -36,6 +41,8 @@ export class ProductAddToQuoteDialogComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
+    this.user$ = this.accountFacade.user$;
+    this.selectedQuoteRequest$ = this.quotingFacade.quoteRequest$;
     this.activeQuoteRequest$ = this.quotingFacade.activeQuoteRequest$;
```
shauke: The `selectedQuoteRequest$` and `activeQuoteRequest$` handling does not work as expected if there is more than one 'NEW' unsubmitted quote request. In that case submitting the quote request in the dialog will not present the "Thank you for your quote" message but switch to the now different active quote request.

The whole handling/distinction of the `selectedQuoteRequest$` and `activeQuoteRequest$` needs to be more clear here (maybe some comments in the code?). Maybe there should not be both within the dialog?

src/app/extensions/quoting/shared/product/product-add-to-quote-dialog/product-add-to-quote-dialog.component.ts
```diff
@@ -48,11 +53,11 @@ export class ProductAddToQuoteDialogComponent implements OnInit, OnDestroy {
...
 
-    this.activeQuoteRequest$
+    // make active quote request to selected
+    this.quotingFacade.activeQuoteRequest$
       .pipe(
-        filter(quoteRequest => !!quoteRequest),
-        distinctUntilKeyChanged('id'),
-        takeUntil(this.destroy$)
+        whenTruthy(),
+        take(1)
```
shauke: `take(1)` instead of `takeUntil(this.destroy$)` solved the problem of not presenting the "Thank you for your quote" message if there is more than one unsubmitted quote request and instead switching to the next unsubmitted quote request.
But with that change the " Create New Request from Quote" in the dialog will create a new quote request but not switch to it in the dialog.

shauke: The commit "removed unnecessarily used activeQuoteRequest$ in favour of selectedQuoteRequest$" should not have changed this behavior and was just introduced to clean up the disturbing usage of two different quote request observables.

PR #130, opened 2020-02-26
fix: change password redirects to login page on success

Sebastian-Haehnlein:
## PR Type

 [x] Bugfix  

change password (forgot password) open login dialog on success 

## What Is the New Behavior?

change password (forgot password) redirects to login page on success 

 [ ] Yes  
 [x] No

### DIFF
--- a/src/app/pages/forgot-password/update-password/update-password.component.ts
+++ b/src/app/pages/forgot-password/update-password/update-password.component.ts
@@ -46,7 +46,7 @@ export class UpdatePasswordComponent implements OnInit, OnDestroy {
         takeUntil(this.destroy$)
       )
       .subscribe(() => {
-        this.router.navigate(['/login']);
+        this.router.navigate(['/login'], { queryParams: { forcePageView: true } });
       });
   }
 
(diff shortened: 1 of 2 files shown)


dhhyi:
> This might fix the behavior but it doesn't repair the original cause. I think it is the `ofRoute('login')` in `user.effects`, that's no longer reached with the login modal. So please have another go at it 😉

Wow, I was under the impression that the password change already logs in the customer, which in my mind would be the only logical thing to do. Turns out I was wrong. So yeah, you're right with the suggested solution. 😅 👍

src/app/pages/forgot-password/update-password/update-password.component.ts
```diff
@@ -46,7 +46,7 @@ export class UpdatePasswordComponent implements OnInit, OnDestroy {
         takeUntil(this.destroy$)
       )
       .subscribe(() => {
-        this.router.navigate(['/login']);
+        this.router.navigate(['/login'], { queryParams: { forcePageView: 'true' } });
```
shauke: I would think a string is not the better choice here. We should just use as simple boolean instead, as is done at other places as well.
This will make the condition in the guard nicer to read too.

PR #180, opened 2020-04-02
feat: added mocks for browser to fix window, document, HTMLElement an…

nickyhaze:
…d navigator issues on server

## PR Type
 [ ] Bugfix  
 [x] Feature  
 [ ] Build-related changes  
 [ ] CI-related changes  
 [x] Other: Server related change

We sometimes use window and prevent issues by using 'isPlatformBrowser'. This works fine, but unfortunately we sometimes face issues with third party dependencies (AgmMarkerCluster for example), becaude they use window without checking if it's defined.

## What Is the New Behavior?
With these browser mocks we prevent any issues regarding window, document, HTMLElement and navigator.

 [ ] Yes  
 [x] No

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -1,3 +1,4 @@
+// tslint:disable:no-string-literal
 // tslint:disable: no-console ish-ordered-imports
 /**
  * *** NOTE ON IMPORTING FROM ANGULAR AND NGUNIVERSAL IN THIS FILE ***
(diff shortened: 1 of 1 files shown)


dhhyi:
@nickyhaze Thank you for your contribution. 👍 

In the best case scenario, we can always use Angular specific adaptions of libraries that are hopefully SSR-ready. I have seen it myself in projects that this is not always the case. However, for the base project, we should strongly stick to the _clean_ approach, so customer projects have the best solution available for customization. Long story short: I would rather see this feature not activated by default. 😅 

Would you be fine with the addition of a commented-out version?

dhhyi:
> Would you be fine with the addition of a commented-out version?

sth like this:
```typescript
// uncomment this block to prevent ssr issues with third-party libraries regarding window, document, HTMLElement and navigator
// tslint:disable-next-line: no-commented-out-code
/*
const domino = require('domino');
const template = fs.readFileSync(join(DIST_FOLDER, 'browser', 'index.html')).toString();
const win = domino.createWindow(template);

// tslint:disable:no-string-literal
global['window'] = win;
global['document'] = win.document;
global['HTMLElement'] = win.HTMLElement;
global['navigator'] = win.navigator;
// tslint:enable:no-string-literal
*/
```

nickyhaze:
Perfect, sounds like a good solution to me!

PR #188, opened 2020-04-10
nginx device type detection

### LINKED ISSUE #171: Server side rendering: deviceType
Hello,

I was wondering if it's possible to detect the device on the server in some sort of way? It's now always 'mobile' by default, is there a reason for this?

Regards,
Nicky Haze - Incentro

dhhyi:
## PR Type

 [x] Feature  

SSR always serves a initial page response for mobile devices. This way some flickering is visible, when accessing the response with desktop browsers.

Issue Number:
Closes #171 

## What Is the New Behavior?

Device type is detected in nginx and forwarded to SSR, so an appropriate response can be rendered and individually cached.

 [ ] Yes  
 [x] No

# Other Information

Builds on top of #167

### DIFF
--- a/nginx/device.conf
+++ b/nginx/device.conf
@@ -1,7 +1,7 @@
 # taken from https://gist.github.com/perusio/1326701
 
 map $http_user_agent $ua_device {
-	default 'pc';
+	default 'desktop';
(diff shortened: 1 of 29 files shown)


nginx/device.conf
```diff
@@ -0,0 +1,7 @@
+# taken from https://gist.github.com/perusio/1326701
+
+map $http_user_agent $ua_device {
+	default 'pc';
```
shauke, line 4: Can we use 'desktop' instead of 'pc' as default? Sounds better, is used in the referenced example too and will include Mac users. ;)

src/environments/environment.model.ts
```diff
@@ -72,6 +72,9 @@ export interface Environment {
   // default viewType used for product listings
   defaultProductListingViewType: ViewType;
 
+  // default device type used for initial page responses
```
shauke, line 75: Listing the 3 options would be nice here I guess.

shauke: Sorry, I missed the fact that this was typed already. No need for additional documentation. :(

nginx/device.conf
dhhyi: Sure, added with breaking change comment. 👍

Issue #181, opened 2020-04-07
Environment variables


nickyhaze:
We’re facing some issues regarding server side rendering and environment variables.
When we do a hard refresh (ctrl+f5) we see the right environment variables (process.env variables) coming from the server. 

On a normal refresh (f5) we see the production variables coming from environment.prod.ts, because we do a prod build and use Docker. We want the application to always check the server variables first, but this doesn’t seem to work like that.

With Docker we should use the same image for each environment with different process.env configuration right?
So if that's the case it should always get the environment variables from the process.env or do we still need to create the angular application using npm build -c test and npm build -c accept?

dhhyi:
Accessing `process.env` ist only possible in the Server-Side Rendering process. All variables there should be put into the NgRx state and then they are transferred via state transfer to the client-side application. This process in described in [Guide - Intershop Progressive Web App 0.18 - Configuration](https://support.intershop.com/kb/index.php/Display/29K337#GuideIntershopProgressiveWebApp0.18-Configuration). To extend it we have added a short guide in you new documentation [here](https://github.com/intershop/intershop-pwa/blob/ebdd59034d1b6c7af9f6982caf729d427179a5f9/docs/guides/propagating-environment-variables.md).

What you are describing as the difference between hard and soft refresh could come from an activated ServiceWorker, yet somehow it does not seem to use the transferred state and therefor falls back to compile-time configurations, normally this should not happen if everything is set up properly.

If you cannot solve this otherwise, you can of course build multiple docker images with different configurations. In the upcoming release there will be a commit that simplifies this: https://github.com/intershop/intershop-pwa/commit/3af4a1de45435eb9fe390ec9d8c027f3314369ad

nickyhaze:
Thank you for the quick reply @dhhyi ! I have placed some logs and I see that it is correctly calling the onServer() function inside ngrx-state-transfer.ts. But it never reaches my log inside the this.transferState.onSerialize callback

Any idea why this is happening?

dhhyi:
> But it never reaches my log inside the this.transferState.onSerialize callback

Where are you checking for these log entries? You won't see them in the browser dev tools, only on the console of the SSR process (`npm run start` or running the docker image). State-Transfer doesn't happen with client-side rendering of the dev-server (`ng serve`).

I just checked. Logs get printed.

dhhyi:
@nickyhaze after some investigation with different settings, I found a constellation where the activated service worker effectively deactivated the state transfer. 

[...]

PR #198, opened 2020-04-19
fix: repair service worker boot-up with correct configuration

dhhyi:
## PR Type

 [x] Bugfix  

Service Worker will boot up with settings from `environment.prod.ts` regardless of what was provided by environment variables, when `index.html` caching was enabled.

Issue Number: 
Relates to #181 

## What Is the New Behavior?

nginx provides a minimal `index.html` response with just the application shell and state transfer containing configuration data.

 [ ] Yes  
 [x] No


dhhyi:
Integrated into #313

PR #218, opened 2020-04-29
refactor: remove basketId from basket service

MaxKless:
## PR Type

## What is the new behavior?

The basket service provides API to supply the basketId, but the shop basically only uses a single basket, so these methods can be refactored.


dhhyi:
Replaced by #400

PR #272, opened 2020-06-03
feat: user management - get roles

Sebastian-Haehnlein:
## PR Type

[*] Feature

## What Is the New Behavior?

extends customer users with roles


projects/organization-management/src/app/store/users/users.actions.ts
```diff
@@ -23,4 +28,31 @@ export class LoadUsersSuccess implements Action {
...
-export type UsersAction = LoadUsers | LoadUsersFail | LoadUsersSuccess;
+export class LoadUsersRoles implements Action {
+  readonly type = UsersActionTypes.LoadUsersRoles;
+  constructor(public payload: { users: User[] }) {}
+}
+
+export class LoadUserRoles implements Action {
+  readonly type = UsersActionTypes.LoadUserRoles;
+  constructor(public payload: { user: User }) {}
+}
```
dhhyi, line 39: For the request, only the user email is used, right? So the payload should only contain the email, not the whole user object.

projects/organization-management/src/app/store/users/users.reducer.ts
```diff
@@ -46,6 +49,16 @@ export function usersReducer(state = initialState, action: UsersAction): UsersSt
         error: undefined,
       };
     }
+
+    case UsersActionTypes.LoadUserRolesSuccess: {
+      const { user, userRoles } = action.payload;
+
+      return {
+        ...usersAdapter.upsertOne({ ...user, roles: userRoles }, state),
```
dhhyi, line 57: Overwriting the complete user data seems unnecessary. Just updating the roles on the user object would suffice.

projects/organization-management/src/app/services/users/users.service.spec.ts
```diff
@@ -15,9 +17,12 @@ describe('Users Service', () => {
     when(apiService.get(anything())).thenReturn(EMPTY);
 
     TestBed.configureTestingModule({
+      imports: [ngrxTesting()],
       providers: [{ provide: ApiService, useFactory: () => instance(apiService) }],
     });
     usersService = TestBed.inject(UsersService);
+    // @ts-ignore
```
dhhyi, line 24: ?

projects/organization-management/src/app/services/users/users.service.ts
```diff
@@ -1,20 +1,40 @@
...
-  constructor(private apiService: ApiService) {}
+  private loggedInCustomer: Customer;
+
+  constructor(private apiService: ApiService, store: Store) {
+    store.pipe(select(getLoggedInCustomer)).subscribe(x => (this.loggedInCustomer = x));
+  }
 
   getUsers() {
-    return this.apiService.get(`customers/-/users`).pipe(
+    return this.apiService.get(`customers/${this.loggedInCustomer.customerNo}/users`).pipe(
```
dhhyi, line 25: I'm a bit confused, why you have to supply the explicit customerNo here. The call was working previously, right? :thinking:

dhhyi:
obsolete

PR #275, opened 2020-06-08
Feature/pwa application type customer resource

SGrueber:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

## What Is the New Behavior?
Switch customer REST resource u dependence of the application and customer type

[ ] Yes
[n] No

### DIFF
--- a/src/app/core/facades/app.facade.ts
+++ b/src/app/core/facades/app.facade.ts
@@ -2,24 +2,25 @@ import { Injectable } from '@angular/core';
 import { NavigationCancel, NavigationEnd, NavigationStart, Router } from '@angular/router';
 import { Store, select } from '@ngrx/store';
 import { combineLatest, merge } from 'rxjs';
-import { filter, map, mapTo, shareReplay, startWith } from 'rxjs/operators';
+import { filter, map, mapTo, shareReplay, startWith, withLatestFrom } from 'rxjs/operators';
 
 import { getAvailableLocales, getCurrentLocale, getDeviceType, getICMBaseURL } from 'ish-core/store/core/configuration';
 import { getGeneralError, getGeneralErrorType } from 'ish-core/store/core/error';
 import { getBreadcrumbData, getHeaderType, getWrapperClass, isStickyHeader } from 'ish-core/store/core/viewconf';
+import { getLoggedInCustomer } from 'ish-core/store/customer/user';
 import { getAllCountries, getCountriesLoading, loadCountries } from 'ish-core/store/general/countries';
 import { getRegionsByCountryCode, loadRegions } from 'ish-core/store/general/regions';
+import { getServerConfigParameter } from 'ish-core/store/general/server-config';
 
 @Injectable({ providedIn: 'root' })
 export class AppFacade {
-  icmBaseUrl: string;
-
   constructor(private store: Store, private router: Router) {
     // tslint:disable-next-line: rxjs-no-ignored-subscribe
     this.routingInProgress$.subscribe();
 
     store.pipe(select(getICMBaseURL)).subscribe(icmBaseUrl => (this.icmBaseUrl = icmBaseUrl));
   }
+  icmBaseUrl: string;
 
   headerType$ = this.store.pipe(select(getHeaderType));
   deviceType$ = this.store.pipe(select(getDeviceType));
@@ -62,6 +63,30 @@ export class AppFacade {
       mapTo(false)
     )
   ).pipe(startWith(true), shareReplay(1));
+
+  /**
+   * selects whether the current application type is 'REST'. If the application type is unknown it returns true
+   */
+  isAppTypeREST$ = this.store.pipe(
+    select(
+      getServerConfigParameter<'intershop.REST' | 'intershop.B2CResponsive' | 'intershop.SMBResponsive'>(
+        'application.applicationType'
+      )
+    ),
+    map(appType => appType === 'intershop.REST' || !appType)
+  );
+
+  customerRestResource$ = this.isAppTypeREST$.pipe(
+    withLatestFrom(this.store.pipe(select(getLoggedInCustomer))),
+    map(([isRest, customer]) => AppFacade.getCustomerRestResource(!customer || customer.isBusinessCustomer, isRest))
+  );
+
+  static getCustomerRestResource(
+    isBusinessCustomer: boolean,
+    isAppTypeREST: boolean
+  ): 'customers' | 'privatecustomers' {
+    return isAppTypeREST && !isBusinessCustomer ? 'privatecustomers' : 'customers';
+  }
 
   countries$() {
     this.store.dispatch(loadCountries());
--- a/src/app/core/models/customer/customer.mapper.ts
+++ b/src/app/core/models/customer/customer.mapper.ts
@@ -6,7 +6,7 @@ export class CustomerMapper {
    * Get User data for the logged in Business Customer.
    */
   static mapLoginData(data: CustomerData): CustomerUserType {
-    return data.type === 'SMBCustomer'
+    return CustomerMapper.isBusinessCustomer(data)
       ? {
           customer: CustomerMapper.fromData(data),
           user: undefined,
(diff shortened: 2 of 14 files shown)


src/app/core/models/customer/customer.mapper.ts
```diff
@@ -48,8 +48,15 @@ export class CustomerMapper {
...
           isBusinessCustomer: false,
         };
   }
+
+  private static isBusinessCustomer(data: CustomerData): boolean {
+    if (data.type === 'PrivateCustomer') {
+      return false;
+    }
+    return true;
+  }
```
dhhyi, line 61: Now would be the perfect time to solve this once and for all: https://github.com/intershop/intershop-pwa/projects/4#card-35562084

Only use the 'isBusinessCustomer' field and remove the 'type' field and also remove this helper method, please :wink:

src/app/core/services/user/user.service.ts
```diff
@@ -209,4 +230,8 @@ export class UserService {
     };
     return this.apiService.post('security/password', data, options);
   }
+
+  private getResource(isBusinessCustomer: boolean): 'customers' | 'privatecustomers' {
+    return this.isAppTypeREST && !isBusinessCustomer ? 'privatecustomers' : 'customers';
+  }
```
dhhyi, line 236: doesn't this duplicate some stuff from the selector output?

src/app/core/services/user/user.service.ts
```diff
@@ -33,8 +37,13 @@ interface CreateBusinessCustomerType extends Customer {
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
-  constructor(private apiService: ApiService) {}
+  private isAppTypeREST: boolean;
+  private customerRestResource = 'customers';
 
+  constructor(private apiService: ApiService, store: Store) {
+    store.pipe(select(isAppTypeREST)).subscribe(isTypeREST => (this.isAppTypeREST = isTypeREST));
```
dhhyi, line 44: I just tried using a facade and I see no issues there. What was the issue when you tried it?

src/app/core/services/address/address.service.ts
```diff
@@ -1,26 +1,32 @@
...
  * The Address Service handles the interaction with the REST API concerning addresses.
  */
 @Injectable({ providedIn: 'root' })
 export class AddressService {
-  constructor(private apiService: ApiService) {}
+  private customerRestResource = 'customers';
+
+  constructor(private apiService: ApiService, store: Store) {
+    store.pipe(select(getCustomerRestResource)).subscribe(restResource => (this.customerRestResource = restResource));
+  }
```
dhhyi, line 21: I think it would be better, if the UserService supplies a method for the customer endpoint and this method is then used within this class

dhhyi: ... also for the other services depending on the specific endpoint. :wink:

src/app/core/services/user/user.service.ts
SGrueber: No, the selector uses the logged in user to determine the rest resource, that's why for the user registration it is not possible to use the selector

src/app/core/services/user/user.service.ts
SGrueber: I got an error at runtime: cannot read property 'subscribe' of null

dhhyi: I tried it without subscribe. As all the calls are RxJS, you can use it with `withLatestFrom` :wink:

src/app/core/services/address/address.service.ts
dhhyi: I have to take this back. As AddressService depends on ApiService via `resolveLinks` nothing would be gained. :cry: I'll have to sleep on it once more.

src/app/core/services/user/user.service.ts
dhhyi: But the surrounding logic can be extracted as it is duplicated. A helper function would be needed that spits out the endpoint depending on a boolean flag.

src/app/core/models/customer/customer.mapper.ts
```diff
@@ -50,4 +50,11 @@ export class CustomerMapper {
           isBusinessCustomer: false,
         };
   }
+
+  private static isBusinessCustomer(data: CustomerData): boolean {
+    if (data.type === 'PrivateCustomer') {
```
shauke, line 55: As discussed, the type information is almost irrelevant. It just tells you what REST resource you were just using. It should not be used for distinction.
The marker for distinction should be the returned `customerType`. This is not yet implemented (see IS-30018) so as a work around the return of a `companyName` property is used as marker for now to distinct BusinessCustomers (has a companyName) from PrivateCustomers.

SGrueber: This condition is right, the distinction is made by the question which rest resource is used
The service decides which rest resource it asks, so mapper doesn't have to take care of this

src/app/core/models/customer/customer.mapper.ts
SGrueber: done in develop

src/app/core/services/user/user.service.ts
dhhyi: yip, like that! 😄 👍

src/app/core/services/user/user.service.ts
```diff
@@ -34,8 +36,10 @@ interface CreateBusinessCustomerType extends Customer {
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
-  constructor(private apiService: ApiService) {}
+  // private isAppTypeREST: boolean;
+  // private customerRestResource = 'customers';
```
dhhyi, line 40: dead code? 🤔

src/app/core/facades/app.facade.ts
```diff
@@ -2,24 +2,30 @@ import { Injectable } from '@angular/core';
...
 export class AppFacade {
-  icmBaseUrl: string;
-
   constructor(private store: Store, private router: Router) {
     // tslint:disable-next-line: rxjs-no-ignored-subscribe
     this.routingInProgress$.subscribe();
 
     store.pipe(select(getICMBaseURL)).subscribe(icmBaseUrl => (this.icmBaseUrl = icmBaseUrl));
   }
+  icmBaseUrl: string;
```
dhhyi, line 28: I think some auto-formatting positioned this here, I'll investigate.. 🤔

src/app/core/services/user/user.service.ts
```diff
@@ -151,11 +182,16 @@ export class UserService {
...
-    if (!customer.isBusinessCustomer) {
-      return this.apiService.put('customers/-/credentials/password', { password, currentPassword });
-    } else {
-      return this.apiService.put('customers/-/users/-/credentials/password', { password, currentPassword });
-    }
+    return this.appFacade.customerRestResource$.pipe(
+      concatMap(restResource =>
+        customer.isBusinessCustomer
+          ? this.apiService.put<void>('customers/-/users/-/credentials/password', { password, currentPassword })
+          : this.apiService.put<void>(`${restResource}/-/credentials/password`, {
```
dhhyi, line 189: The ternary can be moved into the first argument, but I don't know if this is more readable. use your best judgement 😃

src/app/extensions/wishlists/services/wishlist/wishlist.service.spec.ts
```diff
@@ -12,12 +13,20 @@ import { WishlistService } from './wishlist.service';
...
+    appFacade = mock(AppFacade);
+
     TestBed.configureTestingModule({
-      providers: [{ provide: ApiService, useFactory: () => instance(apiServiceMock) }],
+      providers: [
+        { provide: ApiService, useFactory: () => instance(apiServiceMock) },
+        { provide: AppFacade, useFactory: () => instance(appFacade) },
+      ],
     });
+    when(appFacade.customerRestResource$).thenReturn(of('privatecustomers'));
```
dhhyi, line 28: one test for the other case would be nice 😉

src/app/core/facades/app.facade.ts
SGrueber: I leave this as it is

PR #279, opened 2020-06-09
feat: extend budgets to organization users

Sebastian-Haehnlein:
## PR Type

[*] Feature

## What Is the New Behavior?

Add budgets to organization users.


src/app/core/models/user/user.model.ts
```diff
@@ -18,4 +20,6 @@ export interface User {
   // Business User only
   businessPartnerNo?: string;
   department?: string;
+
+  budgets?: UserBudgets;
```
dhhyi, line 24: I think, budgets should be held in an individual store within the user management. Do we really need the values outside of the user management?

dhhyi:
replaced by #309

PR #280, opened 2020-06-10
refactor: remove type from customer model

SGrueber:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

[x] Yes
[ ] No

The customer model has no attribute 'type' any more. Use isBusinessCustomer flag instead to decide whether the customer is a business or private customer.


src/app/core/services/user/user.service.ts
```diff
@@ -72,8 +73,8 @@ export class UserService {
       return throwError('createUser() called without required body data');
     }
 
-    if (!body.customer.type) {
-      return throwError('createUser() called without required customer type (PrivateCustomer/SMBCustomer)');
+    if (body.customer.isBusinessCustomer === undefined) {
+      return throwError('createUser() called without required customer type (isBusinessCustomerRegistration)');
     }
```
dhhyi, line 78: is the feature toggle isBusinessCustomerRegistration dependant on the flag? :thinking:

src/app/core/services/user/user.service.ts
```diff
@@ -112,11 +115,12 @@ export class UserService {
       return throwError('updateUser() called without required body data');
     }
 
-    if (!body.customer.type) {
-      return throwError('updateUser() called without required customer type (PrivateCustomer/SMBCustomer)');
+    if (body.customer.isBusinessCustomer === undefined) {
+      return throwError('updateUser() called without required customer type (isBusinessCustomerRegistration)');
```
dhhyi, line 119: see above

dhhyi: I don't think, we need it any more. 😉

PR #303, opened 2020-06-26
docs: add headless app type migration information

shauke:
Hybrid Approach documentation adaptions for the PWA changes in regards to the new headless application type, plus some helpful other additions.

Note: Needs adaptions once the branch `feature/pwa_application_type` is merged to `develop`.

Wait for `feature/pwa_application_type` to be merged to `develop`.

Also includes upgrade informations for projects staying on RSS.

### DIFF
--- a/docs/concepts/hybrid-approach.md
+++ b/docs/concepts/hybrid-approach.md
@@ -50,6 +50,11 @@ intershop.WebServerSecureURL=https://<nginx>
 
 The server-side rendering process must be started with `SSR_HYBRID=1`.
 
+In addition, the PWA must be run with secure URLs as well.
+This can be achieved with `SSL=1`.
+
+**Only for development environments** it might be necessary to set `TRUST_ICM=1` if the used development ICM is deployed with an insecure certificate.
+
 Also, the **Service Worker must be disabled** for the PWA, as it installs itself into the browser of the client device and takes over the routing process, making it impossible to break out of the PWA and delegate to the ICM.
 
 ### Mapping Table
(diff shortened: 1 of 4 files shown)


docs/concepts/hybrid-approach.md
```diff
@@ -76,6 +81,28 @@ Each entry contains:
...
+
+**Migration Steps to prepare the PWA for the Hybrid Approach with the Responsive Starter Store**
+
+- use a current PWA version
+- revert the following Git commits from Jun 23, 2020
+
+| commit                                                                                                | comment                                                         |
+| ----------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- |
+| [d7504a7](https://github.com/intershop/intershop-pwa/commit/d7504a76c9cf6b944b86f2b4d0160083003b396b) | rename references to 'app_sf_responsive_cm' to 'app_sf_base_cm' |
+| [ddf89a7](https://github.com/intershop/intershop-pwa/commit/ddf89a7f7c0326c5b37e1b2a003db178d96e19c2) | demo content changes - dialogs                                  |
```
dhhyi, line 98: in this commit, the modal for ready-for-shipment is set to lg, is this really part of the migration? 😉

docs/concepts/hybrid-approach.md
```diff
@@ -76,6 +81,28 @@ Each entry contains:
...
+## PWA Adaptions
+
+With version 0.23.0 the PWA was changed to no longer reuse the Responsive Starter Store application types but rather be based upon the newly introduced headless application type for REST Clients - `intershop.REST`.
+This application type would be completely independent of the Responsive Starter Store and is not suited for storefront setups with the Hybrid approach.
+For this reason the PWA needs to be adapted to work together with the Responsive Starter Store again.
+
+**Migration Steps to prepare the PWA for the Hybrid Approach with the Responsive Starter Store**
+
+- use a current PWA version
+- revert the following Git commits from Jun 23, 2020
```
dhhyi, line 93: I'd rather like to try and implement a configurator like https://github.com/intershop/intershop-pwa/tree/develop/schematics/customization/service-worker for the switching between CMS models  rather than telling people to revert commits...

src/hybrid/default-url-mapping-table.ts
```diff
@@ -30,6 +30,11 @@ export const ICM_WEB_URL = '/INTERSHOP/web/WFS/$<channel>/$<lang>/$<application>
 
 /**
  * Mapping table for running PWA and ICM in parallel
+ *
+ * NOTE:
+ * THIS MAPPING TABLE IS JUST AN EXAMPLE IMPLEMENTATION.
+ * IT DOES NOT DEFINE MAPPINGS FOR ALL ROUTES THAT COULD BE HANDLED IN THE PWA OR THE ICM.
+ * IT NEEDS TO BE ADAPTED TO THE NEEDS OF THE PARTICULAR PROJECT.
```
dhhyi, line 37: Maybe the last two lines shouldn't be shouting at developers.

docs/concepts/hybrid-approach.md
shauke: The commit is right, but it might not be a good idea to include the small modal dialog fix in this commit in the first place.
I will need to change that.

docs/concepts/hybrid-approach.md
shauke: I am not certain if this is a scenario that is needed often enough to implement such a configurator?
I will have to think about that. This is just a simple solution for the few cases where it might be needed.

src/hybrid/default-url-mapping-table.ts
shauke: Yes, maybe not.

docs/concepts/hybrid-approach.md
shauke: Removed unneeded and unrelated changes and reduced the set of relevant commits to 2.

docs/concepts/hybrid-approach.md
shauke: We will start with this revert commits approach and think again about a configurator script if there is need and demand for it.

src/hybrid/default-url-mapping-table.ts
shauke: I stopped the continuing shouting.

PR #313, opened 2020-07-03
Prevent Home Page Flickering for SSR

### LINKED ISSUE #311: Flickering CMS content
**Actual Behavior**

During rendering of CMS content it flickers several times, disturbing the user experience. 

**Expected Behavior**

No flickering of CMS content

Open the homepage of PWA and observe the behaviour of CMS content (homepage teasers). Content flickers several times (caused by server-side rendered content being replaced when Angular takes over (first flicker) and then again when CMS REST call is triggered again and content is being updated)

dhhyi:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Issue Number: Closes #311 

## What Is the New Behavior?

- [x] memoized CMS selectors
- [x] ~~CMS functionality into application shell~~
- [x] ~~alternatively: smooth animation~~
- [x] enable Service Worker with loading page
- [x] properly disable Service Worker if disabled (also deletes in browser)
- [x] decide if public deployment should or shouldn't use service worker -> it shouldn't
- [x] disable Service Worker in PWA project by default
- [x] documentation for "Progressive" features

[ ] Yes
[x] No

## Other Information
- adding an animation to the element won't work, as the element holding the animation will be replaced and the animation is stopped.
- adding CMS to the application shell yields no notable change besides completely destroying our current module order


src/app/pages/loading/loading-page.component.html
```diff
@@ -0,0 +1,2 @@
+<ish-loading></ish-loading>
+<p style="height: 400px;"></p>
```
shauke, line 2: If an empty block element is needed to somehow make some room a `<div>` is the better choice since it does not have such a specific semantic meaning as a `<p>` = paragraph.
The used `errorpage` wrapperClass is kind of hacky too, but we can use it for now until we implement a more advanced PWA loading screen. ;)

dhhyi: Yes, let's do that later when we properly use the Service Worker for something _real_. 😄

PR #349, opened 2020-07-28
feat: provide organization interface

skoch-intershop:


projects/organization-management/src/app/models/organization/organization.mapper.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (!data) {
+      throw new Error(`'organizationDocument' is required`);
+    }
+
+    return {
+      id: data.id,
+      name: data.attributes ? data.attributes.name : undefined,
+      description: data.attributes ? data.attributes.description : undefined,
+      authenticationUrl: data.attributes ? data.attributes.authenticationUrl : undefined,
+      nodes: data.relationships.nodes.data.map(identifier => identifier.id),
```
jometzner, line 17: Shouldn't that refer to the node.mapper.ts? So you could work with a node-tree for the given organization?

skoch-intershop: Request contains not enough information to build up an node tree.

projects/organization-management/src/app/models/organization/organization.mapper.ts
```diff
@@ -0,0 +1,21 @@
...
+  static fromData(payload: OrganizationDocument): Organization {
+    const data = payload.data;
+
+    if (!data) {
+      throw new Error(`'organizationDocument' is required`);
+    }
+
+    return {
+      id: data.id,
+      name: data.attributes ? data.attributes.name : undefined,
```
jometzner, line 14: Please use safe navigation with `?.` operator

jometzner: ALso `id` and `name` are both mandatory attributes according to the API description

PR #356, opened 2020-08-04
feat: styled toast messages for success and error messages

dhhyi:
## PR Type

[x] Feature

- error messages are mostly displayed inline
- some success messages are displayed as toasts, but only in the top right corner

## What Is the New Behavior?

- toast messages used and styled more consistently

[ ] Yes
[x] No

## TODO:
- [x] error messages should not fade out
- [x] go to URLs provided in toast message
- [x] adapt cypress tests
- [x] adapt font size
- [x] add cross action icon for error messages (to indicate that you can click it away)
- [x] In which cases should it be switched off? (e.g. errors in modals, info messages) tbd.
- [x] mobile layout (bottom?)

To be discussed:
- [ ] consistent toast positioning when sticky header is disabled (under header when header is in view, top of screen otherwise)
there is no simple solution, maybe postponed after header rework

Toasts are not used (yet):
- in the checkout because of the divergent header
- for information messages - consider to create a general info-message component in future for it
- success messages in modals are not replaced, e.g. add product to wish list/order template

VD Template:
toast

### DIFF
--- a/e2e/cypress/integration/pages/account/registration.page.ts
+++ b/e2e/cypress/integration/pages/account/registration.page.ts
@@ -105,6 +105,6 @@ export class RegistrationPage {
   }
 
   get errorText() {
-    return cy.get('div.alert');
+    return cy.get('.toast-container .toast-error');
   }
 }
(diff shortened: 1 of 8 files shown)


M-Behr (changes requested):
Please have a look at the tablet view. I think the message has to be a little bit larger to fill the whole 
content area.
Toast_Tablet

M-Behr (approved):

PR #366, opened 2020-08-26
Feature/company structure ngx treeview

jometzner:
Adds tree display based on [ngx-treeview](https://github.com/leovo2708/ngx-treeview) library. We'll very likely remove the dependency later because of lodash dependency. But we needed to have some tree component in place to showcase where we are. The task to select a better option is in the team backlog: ISREST-1068

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21882,6 +21882,14 @@
         "tslib": "^2.0.0"
       }
     },
+    "ngx-treeview": {
+      "version": "10.0.2",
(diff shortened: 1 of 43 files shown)


jometzner:
```
WARNING in /home/user/work/intershop-pwa/node_modules/ngx-treeview/__ivy_ngcc__/fesm2015/ngx-treeview.js depends on 'lodash'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```

PR #371, opened 2020-08-31
refactor: allow numbers as loading for ngrx reducer helpers

dhhyi:
## PR Type

`loading` in reducers is boolean by default

## What Is the New Behavior?

`loading` can also be number for multiple parallel loading actions.

[ ] Yes
[x] No


shauke (approved):
I would keep "unsetLoadingAndError" since I have no better idea myself.

PR #373, opened 2020-09-01
perf: remove product views from entities for improved memoization

dhhyi:
## PR Type

Entities containing line items with products (Order, Basket, Quotes, QuoteRequests) project ProductViews in selectors for easy access. Most of the access is however abstracted with LineItemListComponent.

## What Is the New Behavior?

ProductViews are no longer part of line item selectors. That way the selector output memoization improves. Relevant components fetch products via facade if necessary.

[ ] Yes
[x] No


shauke (approved):
Nice code simplifications. 👍

PR #375, opened 2020-09-03
refactor: node script to clean up localization files

suschneider:
## PR Type

The localization files are copies of the Responsive Starter Store localization files and contain many unused localizations.

## What Is the New Behavior?

The localization files only contain the localizations used. A script is provided to regularly clean up the localization files.

[ ] Yes
[x] No

## TODO:
- [x] write node script that cleans up localization files
- [x] implement pattern check for not explicitly used localization keys (e.g. dynamically generated keys, keys used in REST) 
- [x] introduce a consistent localization key ordering in the cleaned localization files
- [x] expand to other languages (french, german)
- [x] implement error handling and logging for script
- [x] integrate script in the build process
- [x] documentation


scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
+const fs = require('fs');
+const glob = require('glob');
+
+const localizationFile_en = 'src/assets/i18n/en_US.json';
+const localizationFile_fr = 'src/assets/i18n/fr_FR.json';
+const localizationFile_de = 'src/assets/i18n/de_DE.json';
```
dhhyi, line 6: I'd rather not hard-code this, but make it generic. I guess we can expect `en_US.json` to always be there, but other files should be discovered automatically.

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+
+// regular expressions for patterns of not explicit used localization keys (dynamic created keys, error keys from rest calls)
+// new patterns have to be added here
+const patterns = ['account.login..*.message', '.*.?error..*'];
+const regEx = new RegExp(patterns.join('|'), 'i');
```
dhhyi, line 11: `const regEx = /account\.login\..*\.message|.*\.error\..*/i`

If a dot is unescaped, it matches any character. In this case we want to explicitly match dots.

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+    localizationsFoundOrdered[key] = localizationsFound[key];
+  });
+
+// write found localizations into file
+fs.writeFileSync(localizationFile_en, JSON.stringify(localizationsFoundOrdered, null, 2));
+
+// create cleaned localization files for other languages
+Object.keys(localizationsFoundOrdered).forEach(localizationKey => {
+  localizationsFoundOrdered[localizationKey] = localizations_fr[localizationKey];
+});
```
dhhyi, line 60: you could use lodash [`pick`](https://lodash.com/docs/4.17.15#pick) for this

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+  .filter(v => v.match(regEx))
+  .map(localizationKey => {
+    localizationsFound[localizationKey] = localizations[localizationKey];
+    delete localizations[localizationKey];
+  });
+
+filesToBeSearched.forEach(filePath => {
+  const fileContent = fs.readFileSync(filePath);
+  for (var localizationKey in localizations) {
+    if (fileContent.indexOf(localizationKey) > -1) {
```
dhhyi, line 39: `fileContent.includes(localizationKey)`

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+const localizationFile_de = 'src/assets/i18n/de_DE.json';
+
+// regular expressions for patterns of not explicit used localization keys (dynamic created keys, error keys from rest calls)
+// new patterns have to be added here
+const patterns = ['account.login..*.message', '.*.?error..*'];
+const regEx = new RegExp(patterns.join('|'), 'i');
+
+var filesToBeSearched = [];
+var localizationsFound = {};
+var localizationsFoundOrdered = {};
```
dhhyi, line 15: For increased readability please declare variables, when you first use them.

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+fs.writeFileSync(localizationFile_en, JSON.stringify(localizationsFoundOrdered, null, 2));
+
+// create cleaned localization files for other languages
+Object.keys(localizationsFoundOrdered).forEach(localizationKey => {
+  localizationsFoundOrdered[localizationKey] = localizations_fr[localizationKey];
+});
+fs.writeFileSync(localizationFile_fr, JSON.stringify(localizationsFoundOrdered, null, 2));
+Object.keys(localizationsFoundOrdered).forEach(localizationKey => {
+  localizationsFoundOrdered[localizationKey] = localizations_de[localizationKey];
+});
```
dhhyi, line 64: Consider handling `localizationsFoundOrdered` as a `string[]`, then you can remove some of the `Object.keys` invocations and instead just iterate over it.

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+// store localizations from default localization file in an object
+var localizations = JSON.parse(fs.readFileSync(localizationFile_en, 'utf8'));
+
+var localizations_fr = JSON.parse(fs.readFileSync(localizationFile_fr, 'utf8'));
+var localizations_de = JSON.parse(fs.readFileSync(localizationFile_de, 'utf8'));
+
+// go through directory recursively and find files to be searched
+glob.sync('{src,projects}/**/!(*.spec).{ts,html}').forEach(filePath => {
+  filesToBeSearched.push(filePath);
+});
```
dhhyi, line 26: `const filesToBeSearched = glob.sync ...`

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,65 @@
...
+
+// add not explicit used localization keys
+Object.keys(localizations)
+  .filter(v => v.match(regEx))
```
dhhyi, line 30: String.[match](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/match) returns an array with matches. In this case I think we just want to have a boolean, if the key matches the regular expression. RegExp.[test](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) is more suitable for this.

dhhyi (changes requested):

scripts/clean-up-localizations.js
dhhyi: ... but then the sorting could be broken. Could you test this, please? 😃

scripts/clean-up-localizations.js
suschneider: With the use of lodash here this comment is done.

shauke (dismissed):
Nicely done script. 👍 

The only issue we still need to resolve is the failing "Check No Changes" test after "npm run check" with the changed localization files on the CI system. A line ending issue?

dhhyi (changes requested):
For projects there should be two commits on develop: one commit should add the script and one should modify all localization files, so the latter can be potentially skipped while migrating.
Can you re-run the script on the unmodified translation files again? I'm getting strange behaviors...
Also, please add a note to the migration guide.

scripts/clean-up-localizations.js
```diff
@@ -0,0 +1,75 @@
...
+localizationFiles_lang.forEach(file => {
+  console.log('\nClean up file', file);
+  var localizations_lang = JSON.parse(fs.readFileSync(file, 'utf8'));
+  // find missing localization keys
+  _.difference(Object.keys(localizationsFoundOrdered), Object.keys(localizations_lang)).forEach(key => {
+    console.log('\x1b[33m%s\x1b[0m', `  Localization key ${key} not found in ${file}`);
+  });
+  fs.writeFileSync(
+    file,
+    JSON.stringify(_.pick(localizations_lang, Object.keys(localizationsFoundOrdered)), null, 2) + '\n'
```
dhhyi, line 73: instead of adding the newline, it might just be better, to run prettier on the changed files: `execSync('npx prettier --write ' + path);`

shauke:
The commits are not cleaned up for merging yet, but the one separate commit with only the changes to the localization files is already prepared with the same intention as mentioned by you "so the latter can potentially be skipped while migrating".
Good point in regards to the note for the migration guide though.

dhhyi:
I'm still getting a strange behavior when running the script on develop's files. It tells me the key `account.user.update_profile.heading` is missing in French and German, but it's not missing on this branch? Did somebody try to sneak in missing translations in the cleanup commit? This should be an additional commit, right? 😉

.github/workflows/windows.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: Display Changes
         if: failure()
-        run: git diff
+        run: git diff-files && git diff
```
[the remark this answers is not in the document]
shauke: This is currently for testing and failure analysis. 'git diff' is not returning anything with the current problem so I added 'git diff-files' to see at least which files do have a problem and lead to failure in the preceding check. Unfortunately I cannot see what diffs there actually are in the files. This could be reverted once we resolve the problem but it might be helpful information in some cases.

scripts/clean-up-localizations.js
shauke: Yes, this is more or less an act of desperation since we assume - but cannot actually prove in the CI pipeline - that this missing new line at the file ending might be the problem. Since it does not resolve the problem we will remove this hack again. CI runs with localization files commited without a newline at the end did not resolve the problem either.
The idea with an additional prettier run sounds good. I will test that. Or maybe I will move the localization check in front of the formatting run within 'npm run check'? Thanks for the idea.

.github/workflows/windows.yml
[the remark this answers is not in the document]
dhhyi: I see. I guess you could permanently replace the `--quiet` flag from the previous command with `--exit-code` for this: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---quiet

shauke:
Yes that is right. ;)
For an optional commit we need to add the missing `account.user.update_profile.heading` keys for german and french in a separate commit. If this is your strange behavior then this is the explaination. Good call.

scripts/clean-up-localizations.js
dhhyi: If you change the order in `npm run check`, then the script itself will not produce proper output when run individually. I would prefer the manual run in the script itself. We do this in many scripts, and it's running quite fast when pointed to individual files 😉

.github/workflows/windows.yml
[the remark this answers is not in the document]
shauke: Changed to `run: git diff-files --exit-code`.

scripts/clean-up-localizations.js
shauke: Added the prettier run to the script. Unfortunately it does not resolve the problem with the diffs in the localization files. :(

shauke:
Added a separate commit for the added localizations. Added migration information.

scripts/clean-up-localizations.js
dhhyi: exchanged `git diff-files` with `git diff`.

dhhyi (dismissed):

shauke (approved):

PR #376, opened 2020-09-04
fix: repair loading of lazy extension components

dhhyi:
## PR Type

[x] Bugfix
[x] Build-related changes

Lazy components and extension stores are loaded without relation to their extension modules.

## What Is the New Behavior?

Fixed.

[x] Yes
[ ] No


shauke (approved):

shauke:
After this change lazy loading modules became lazy loaded at all again. Before they where part of the main bundle which was not the intended way.

PR #397, opened 2020-10-02
fix: display and edit the b2b user's active status

SGrueber:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

It is not possible to see or change the active status of a user in the b2b user management.

## What Is the New Behavior?
You can see the user detail page, whether a b2b user is inactive. You can activate/deactivate a b2b user on b2b user creation and edit page.

[ ] Yes
[x] No


M-Behr (approved):

iwiederhold (dismissed):

projects/organization-management/src/app/models/b2b-user/b2b-user.mapper.ts
```diff
@@ -17,6 +17,7 @@ export class B2bUserMapper {
         firstName: AttributeHelper.getAttributeValueByAttributeName(e.attributes, 'firstName'),
         lastName: AttributeHelper.getAttributeValueByAttributeName(e.attributes, 'lastName'),
         roleIDs: AttributeHelper.getAttributeValueByAttributeName(e.attributes, 'roleIDs'),
+        // ToDo: #IS-31051: map the active flag
```
dhhyi, line 20: You can already insert the code and remove the TODO. It will just start working, when the fix arrives 😄 
But, please add a note as "known limitation" in the PR description.

src/app/core/models/user/user.interface.ts
```diff
@@ -24,4 +24,6 @@ export interface UserData {
 
   businessPartnerNo?: string;
   department?: string;
+
+  active?: boolean;
```
dhhyi, line 28: This property arrives only in B2B and actually does not carry additional information for the current user (because it has to be active there). Can we instead use proper B2bUser Data types? I'll commit a fixup as suggestion, please remove it if you don't agree 😉

dhhyi (changes requested):

dhhyi (dismissed):

dhhyi (approved):

PR #400, opened 2020-10-05
Refactor/service remove basket id 2

MaxKless:
## PR Type

## Changes

The basket service is able to access the current basket via the "baskets/current" route and thus does not need a valid **basketId** in its calls.
This pull request removes the **basketId** parameter from all unnecessary service and effect calls. 
In addition, the addItemsToBasket and createBasketBeforeAddItemsToBasket effects are combined for more straightforward behaviour.

### DIFF
--- a/src/app/core/services/basket/basket.service.spec.ts
+++ b/src/app/core/services/basket/basket.service.spec.ts
@@ -52,11 +52,11 @@ describe('Basket Service', () => {
   });
 
   it("should get basket data when 'getBasket' is called", done => {
-    when(apiService.get(`baskets/${basketMockData.data.id}`, anything())).thenReturn(of(basketMockData));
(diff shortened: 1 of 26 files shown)


src/app/core/services/basket/basket.service.spec.ts
```diff
@@ -52,11 +52,11 @@ describe('Basket Service', () => {
   });
 
   it("should get basket data when 'getBasket' is called", done => {
```
SGrueber, line 54: There are still many unnecessary occurrences left:
- payment.service
- quoting.service
- actions: - createOrder, addItemsToBasket, quoting Actions

pls. check also other files again

SGrueber (changes requested):

SGrueber (approved):

PR #418, opened 2020-10-13
fix: adjust quote list item padding

### LINKED ISSUE #291: Quote overview page: Different padding in table cells
**Actual Behavior**

The quoting table cells were implemented with a different padding. This styling is inconsistent to other tables, e.g. order templates.

**Expected Behavior**

Use the same styling for all tables in my account. Make sure that the same table component & styles are used. See other places, e.g. order templates, order history, my account overview.

1. Add a quote request
2. Go to my account area
3. Click on 'Quotes'
4. See table

Quotes:
Quotes Table

Order Templates:
Order Templates Table

Tested Version: PWA 0.20.0

MaxKless:
## PR Type

bugfix

quote list item padding is wrong

Issue Number: Closes #291

## What Is the New Behavior?

quote list item padding is consistent with other account tables

No

### DIFF
--- a/src/app/extensions/quoting/pages/quote-list/quote-list/quote-list.component.html
+++ b/src/app/extensions/quoting/pages/quote-list/quote-list/quote-list.component.html
@@ -14,17 +14,17 @@
       </thead>
 
       <tr *ngFor="let quote of quotes">
-        <td>
+        <td class="px-0">
           <a [routerLink]="[quote.id]">{{ quote.number }}</a>
         </td>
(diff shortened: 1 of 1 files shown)


M-Behr (approved):
The padding bug is fixed.
Please think about using the list instead of the table everywhere in my account. The list is used in order history, order templates and user management. The table is used in quotes. The phone view is slightly different between these components. Nevertheless I would agree to change the quote table to a quote list.

PR #421, opened 2020-10-14
docs: improve configuration documentation e.g. with code examples

shauke:
configuration documentation additions

### DIFF
--- a/docs/concepts/configuration.md
+++ b/docs/concepts/configuration.md
@@ -51,8 +51,36 @@ As can be seen here, only build-time and deploy-time configuration parameter can
 
 ### Node.js Environment Variables
 
-When running the application in Angular Universal mode within a _Node.js_ environment, we can additionally access the process environment variables via _process.env._ This method provides a way to configure the application at deploy time, e.g., when using docker images.
-Configuration can then be consumed and passed to the client side via means of state transfer.
+When running the application in Angular Universal mode within a _Node.js_ environment, we can additionally access the process environment variables via _process.env_.
+This method provides a way to configure the application at deploy time, e.g., when using docker images.
+Configuration can then be consumed and passed to the client side via state transfer using Angular's [TransferState](https://angular.io/api/platform-browser/TransferState).
+
+To introduce a new `TransferState` key it should be added to the [`state-keys.ts`](../../src/app/core/configurations/state-keys.ts).
+
+```typescript
+export const NEW_KEY = makeStateKey<string>('newKey');
+```
+
+The actual transfer is handled by the [`app.server.module.ts`](../../src/app/app.server.module.ts) where the mapping of the environment variable is done.
+
+Accessing the transferred state in a service or a component is pretty straight forward:
+
+```typescript
+import { isPlatformBrowser } from '@angular/common';
+import { PLATFORM_ID } from '@angular/core';
+import { NEW_KEY } from 'ish-core/configurations/state-keys';
+
+newKey string;
+
+constructor(
+  @Inject(PLATFORM_ID) private platformId: string,
+  private transferState: TransferState,
+) {}
+
+if (isPlatformBrowser(this.platformId)) {
+  this.newKey = this.transferState.get<string>(NEW_KEY, 'default value');
+}
+```
 
 ### NgRx Configuration State
 
@@ -82,6 +110,8 @@ The most common way of supplying them can be implemented by using Angular CLI en
 
 An example for this kind of settings are breakpoint settings for the different device classes of the application touch points.
 
+Deployment settings can also be set by using `TransferState` to use environment variables of the deployment for the configuration of the application.
+
 ### Runtime Settings
 
 The most flexible kind of settings, which can also change when the application runs, are runtime settings.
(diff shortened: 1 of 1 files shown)


docs/concepts/configuration.md
```diff
@@ -51,8 +51,45 @@ As can be seen here, only build-time and deploy-time configuration parameter can
...
+) {}
+
+const newKey = this.transferState.get<string>(NEW_KEY, 'default value');
+```
+
+With the state transfer in place the `NEW_KEY` could for example be set as a docker image environment variable like this.
+
+```
+docker run -d -p <port>:4200 -e NEW_KEY=foo_bar --name <container_name> <pwa_image_name>
+```
```
dhhyi, line 92: Please, remove the example call. Everybody knows how to set environment variables with docker. 😉 Besides, it's a different method for docker-compose or kubernetes, or...

docs/concepts/configuration.md
```diff
@@ -51,8 +51,45 @@ As can be seen here, only build-time and deploy-time configuration parameter can
...
+Accessing the transferred state in a service or a component is pretty straight forward.
+
+```typescript
+import { NEW_KEY } from 'ish-core/configurations/state-keys';
+
+constructor(
+  private transferState: TransferState,
+) {}
+
+const newKey = this.transferState.get<string>(NEW_KEY, 'default value');
```
dhhyi, line 85: Shouldn't this be wrapped by `isPlatformBrowser(this.platformId)`?

docs/concepts/configuration.md
```diff
@@ -82,6 +119,8 @@ The most common way of supplying them can be implemented by using Angular CLI en
 
 An example for this kind of settings are breakpoint settings for the different device classes of the application touch points.
 
+Another way would be the usage of `TransferState` to use environment variables of the deployment for the configuration of the application.
```
dhhyi, line 122: "Another way" refers to the sentence before the previous. Please re-add context. i.e. "Deployment settings can also be ..."

docs/concepts/configuration.md
```diff
@@ -92,6 +131,26 @@ Nevertheless, default values can be provided by environment files and can later
...
+
+```typescript
+import { StatePropertiesService } from 'ish-core/utils/state-transfer/state-properties.service';
+
+  constructor(
+    private stateProperties: StatePropertiesService,
+  ) {}
+
+this.stateProperties.getStateOrEnvOrDefault<string>('PROPERTY_KEY', 'propertyKey')
+```
```
dhhyi, line 144: please link to method in the code instead of supplying snippet. Maintainability above all!

docs/concepts/configuration.md
```diff
@@ -92,6 +131,26 @@ Nevertheless, default values can be provided by environment files and can later
...
+
+### Configurations REST Resource
+
+ICM provides a Configurations REST resource - `/configurations` -
+that is supposed to provide all relevant runtime configurations that can be defined in the ICM BackOffice and might be needed to configure a REST client as well.
```
dhhyi, line 149: ~~might~~ I don't think this is optional

docs/concepts/configuration.md
```diff
@@ -92,6 +131,26 @@ Nevertheless, default values can be provided by environment files and can later
...
+This includes service configurations, locales, basket preferences, etc.
+
+The ICM configurations information can be accessed through the `getServerConfigParameter<T>(path)` selector.
```
dhhyi, line 152: link to code

docs/concepts/configuration.md
```diff
@@ -51,8 +51,45 @@ As can be seen here, only build-time and deploy-time configuration parameter can
...
+
+```typescript
+import { NEW_KEY } from 'ish-core/configurations/state-keys';
+
+export class AppServerModule {
+  constructor(transferState: TransferState) {
+    transferState.set(NEW_KEY, process.env.NEW_KEY);
+  }
+}
+```
```
dhhyi, line 74: please link to official TransferState class instead of supplying a code snippet. Linking to `app.server.module.ts` would also be fine, so readers can look it up.

dhhyi (changes requested):

docs/concepts/configuration.md
shauke: I do not see that this is a general requirement https://angular.io/api/platform-browser/TransferState.
The `get` method should work with the `ServerTransferStateModule` as well.
It might be useful in many situations to wrap it but it probably should not be a general recommendation?

docs/concepts/configuration.md
shauke: You would prefer a link to https://github.com/intershop/intershop-pwa/blob/develop/src/app/core/utils/state-transfer/state-properties.service.ts instead of this code snipped on how to actually use it?

docs/concepts/configuration.md
shauke: I believe the 3 code snippets are helpful since they make it more obvious how the whole TransferState works together and what places need to be adapted. I think this would be less clear if I would remove one piece of it and leave this part to the reader to figure it out?

shauke: I guess we should consider helping someone with this documentation that has not yet fully understood the underlying concept or would be able to get the usage just from reading https://angular.io/api/platform-browser/TransferState.

docs/concepts/configuration.md
dhhyi: You're right, it is not a general requirement, but it should be one for the PWA. If on the server a set comes before the get, then you'll always have a value and `'default value'` is obsolete. The default value would then only be used with webpack-dev-server, which is not a production environment.

docs/concepts/configuration.md
dhhyi: Yes, if you don't think the method is properly (and minimally) documented, please improve it there 😉

docs/concepts/configuration.md
dhhyi: For this, there is quite a variety of blog articles that cover this topic. And as we are not doing something entirely custom to the PWA, the developers should read about this on the web.

dhhyi (approved):

marschmidt89 (approved):

PR #437, opened 2020-10-23
feat(schematics): lazy-components for projects

dhhyi:
## PR Type

[x] Feature

lazy-component schematic works only for extensions.

## What Is the New Behavior?

lazy-component schematic also works for projects.

[ ] Yes
[x] No

## TODO:
- [x] adapt schematic
- [x] add test for project
- [x] create exports module if it does not exist
- [x] export lazy component in index.ts for tests?
- [x] adapt tslint rules

### DIFF
--- a/schematics/src/lazy-component/factory.ts
+++ b/schematics/src/lazy-component/factory.ts
@@ -182,38 +182,6 @@ export function createLazyComponent(options: Options): Rule {
       )
     );
 
-    // if (exportsModuleExists) {
-    //   operations.push(
-    //     mergeWith(
-    //       apply(url(exportsModulePath), [
-    //         forEach(() => {
-    //           host.overwrite(exportsModulePath, exportsModuleContent);
-    //           // tslint:disable-next-line: no-null-keyword
-    //           return null;
-    //         }),
-    //       ])
-    //     )
-    //   );
-    //   operations.push(
-    //     addDeclarationToNgModule({
-    //       ...options,
-    //       module: findModuleFromOptions(host, { name: `${classify(exportsModuleName)}Exports` }),
-    //     })
-    //   );
-    //   // addDeclarationToModule(
-    //   //   readIntoSourceFile(host, exportsModulePath),
-    //   //   exportsModulePath,
-    //   //   classify(options.name),
-    //   //   componentPath
-    //   // );
-    //   // operations.push(
-    //   //   addExportToNgModule({
-    //   //     ...options,
-    //   //     module: findModuleFromOptions(host, { ...options, name: exportsModuleName }),
-    //   //   })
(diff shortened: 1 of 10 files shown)


schematics/src/utils/registration.ts
```diff
@@ -180,6 +185,17 @@ export function addImportToFile(options: { module?: string; artifactName?: strin
   };
 }
 
+export function addLazyExportToBarrelFile(options: { path?: string; name?: string; artifactName?: string }): Rule {
```
dhhyi, line 188: This rule factory does not explicitly just add "lazy" components, right? So the name should just be `addExportToBarrelFile`.

schematics/src/utils/registration.ts
```diff
@@ -180,6 +185,17 @@ export function addImportToFile(options: { module?: string; artifactName?: strin
...
+  const barrelFile = `${options.path}/index.ts`;
+  return host => {
+    if (!tsquery(readIntoSourceFile(host, barrelFile), `Identifier[name=${options.artifactName}]`).length) {
+      const exportRecorder = host.beginUpdate(barrelFile);
+      insertExport(exportRecorder, options.artifactName, `./${options.name}/${options.name}.component`);
```
dhhyi, line 193: for this rule the option `moduleImportPath` seems to be set correctly. Could you use it like the `insertImport` in the previous helper `addImportToFile` instead of combining the path directly?

schematics/src/lazy-component/factory.ts
```diff
@@ -108,12 +121,28 @@ export function createLazyComponent(options: Options): Rule {
...
     if (!options.ci) {
+      if (!exportsModuleExists) {
+        operations.push(
+          schematic('module', {
+            ...options,
+            name: exportsModuleName,
+            flat: true,
+          })
+        );
+        operations.push(updateModule(options));
```
dhhyi, line 138: The new module should also be added to the exports barrel file.

schematics/src/lazy-component/factory.ts
```diff
@@ -108,12 +121,28 @@ export function createLazyComponent(options: Options): Rule {
         }
       }
     }
+    const exportsModuleName = `${isProject ? projectName : extension}-exports`;
+    const exportsModuleExists = host.exists(`${options.path}/${exportsModuleName}.module.ts`);
 
     const operations = [];
 
     if (!options.ci) {
+      if (!exportsModuleExists) {
```
dhhyi, line 130: In the same fashion: if the exports folder does not contain a `.gitignore` with `/lazy**`, it should be added, because generated lazy components should not be committed to the source repository.

MaxKless (approved):

PR #458, opened 2020-11-12
fix: display only one validation error for a form element at once

SGrueber:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Multiple validation error messages are shown for one form element, 
e.g. if an email address is required 2 errors are shown, if the field is empty

Issue Number: Closes #
see ISREST-1088

## What Is the New Behavior?
If a form element has several validators, only the 1st validation message is shown

[ ] Yes
[x] No

### DIFF
--- a/src/app/shared/forms/components/form-control-feedback/form-control-feedback.component.html
+++ b/src/app/shared/forms/components/form-control-feedback/form-control-feedback.component.html
@@ -1,5 +1,7 @@
 <ng-container *ngIf="control.dirty && control.validator">
   <fa-icon class="form-control-feedback" *ngIf="this.control.valid; else invalid" [icon]="['fas', 'check']"></fa-icon>
   <ng-template #invalid> <fa-icon class="form-control-feedback" [icon]="['fas', 'times']"></fa-icon> </ng-template>
-  <small *ngFor="let e$ of errors" class="form-text"> {{ e$ | async }} </small>
+  <small *ngFor="let error of errors | slice: 0:1" class="form-text">
+    {{ error | translate }}
+  </small>
 </ng-container>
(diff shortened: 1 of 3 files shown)


src/app/shared/forms/components/form-control-feedback/form-control-feedback.component.html
```diff
@@ -1,5 +1,7 @@
 <ng-container *ngIf="control.dirty && control.validator">
   <fa-icon class="form-control-feedback" *ngIf="this.control.valid; else invalid" [icon]="['fas', 'check']"></fa-icon>
   <ng-template #invalid> <fa-icon class="form-control-feedback" [icon]="['fas', 'times']"></fa-icon> </ng-template>
-  <small *ngFor="let e$ of errors" class="form-text"> {{ e$ | async }} </small>
+  <ng-container *ngFor="let e$ of errors; let i = index">
+    <small *ngIf="i === 0" class="form-text">{{ e$ | async }}</small>
+  </ng-container>
```
dhhyi, line 6: ` | slice:0:1`

dhhyi (changes requested):

src/app/shared/forms/components/form-control-feedback/form-control-feedback.component.html
dhhyi: I don't quite understand why `async` is used on `e$` here. It should be the errors array that should be subscribed to.

SGrueber:
I understand the async at e$ neither, but it is only working that way, all other solutions gave compiler errors

dhhyi:
> I understand the async at e$ neither, but it is only working that way, all other solutions gave compiler errors

ah yes, I misread the declaration.
I think you can safely use translate.instant here to get rid of subscriptions.

dhhyi:
> > I understand the async at e$ neither, but it is only working that way, all other solutions gave compiler errors
> 
> ah yes, I misread the declaration.
> I think you can safely use translate.instant here to get rid of subscriptions.

Or even better: don't do the translation in the component and do it on the template with translate pipe.

dhhyi (approved):
👍

PR #459, opened 2020-11-12
fix: prevent an error if language switched on order details page 

SGrueber:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

After the user changed the language on the order details page an error page is displayed. - This problem only occurs in case of server-side rendering with Angular Universal.

see also IS-IS-31347

## What Is the New Behavior?
The order detail page is displayed in the requested language.

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/store/customer/orders/orders.effects.ts
+++ b/src/app/core/store/customer/orders/orders.effects.ts
@@ -1,11 +1,12 @@
-import { Injectable } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
+import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
 import { Router } from '@angular/router';
 import { Actions, createEffect, ofType } from '@ngrx/effects';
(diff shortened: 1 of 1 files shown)


src/app/core/store/customer/orders/orders.effects.ts
```diff
@@ -163,11 +165,16 @@ export class OrdersEffects {
...
+    iif(
+      () => isPlatformBrowser(this.platformId),
+
+      this.actions$.pipe(
+        ofType(selectOrder),
+        mapToPayloadProperty('orderId'),
+        whenTruthy(),
+
+        map(orderId => loadOrder({ orderId }))
+      )
```
shauke, line 177: Looks good once you get rid of the two unnecessary line breaks. ;)

shauke (changes requested):

dhhyi (approved):
LGTM 👍

shauke (approved):

PR #470, opened 2020-11-26
run cypress tests with ngrx runtime checks

MaxKless:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[x] CI-related changes
[ ] Other: 

Currently, e2e tests are not run with [ngrx runtime](https://ngrx.io/guide/store/configuration/runtime-checks) checks enabled.

## What Is the New Behavior?

Now, they will be run with runtime checks. Tested to fail in case of violation: 
image

[ ] Yes
[x] No

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -18,7 +18,7 @@
       "prefix": "ish",
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:browser",
+          "builder": "@angular-builders/custom-webpack:browser",
           "options": {
             "outputPath": "dist/browser",
             "index": "src/index.html",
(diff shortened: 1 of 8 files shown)


dhhyi (changes requested):
Wouldn't it be better to just run the cypress tests on an environment with production=false instead of introducing another variable?
Please also have a look at https://medium.com/agilix/angular-and-cypress-data-cy-attributes-d698c01df062. There a cli builder is introduced which could be used for removing "data-testing" attributes and therefor further minify our huge page responses for production.

dhhyi:
replaced by #521

PR #478, opened 2020-12-03
chore: fixes for SSO with Auth0 for B2C 

dhhyi:
## TODOs

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -95,6 +95,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         test: ['mock', 'b2c', 'b2b']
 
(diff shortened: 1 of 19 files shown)


M-Behr (changes requested):
Overall, it works well.
I am not sure if these points are fixable:
- [ ] Click "Register" or "Sign in" => Redirect to Auth0 => use browser back button => never-ending PWA loading page
- [ ] Language always changes to English after logging in or out, also Auth0 Login is English after changing the language => Is this not submitted?
- [ ] No e-mails received (e.g. Order Confirmation/Contact Us confirmation)

Open points:
- Login in Checkout
- Editable profile settings=> This was not nice in the backoffice: Every time I wanted to update something in the user profile, I also had to edit the first name, because there was always the e-mail in the input and otherwise errors were displayed => Do we really want that for the PWA, too?
- Active flag is ignored

M-Behr (approved):

MaxKless (approved):

PR #484, opened 2020-12-08
chore: run jest precommit sequentially

dhhyi:
## PR Type

[x] Build-related changes

- The pre-commit hook running jest sometimes hangs if certain TestBed declarations are not properly configured.
- Jest in pre-commit uses all available cores for the test run which can overwhelm developer machines.

## What Is the New Behavior?

Jest is run sequentially in pre-commit. This will also cause it to run significantly slower every time.

[ ] Yes
[x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     ],
     "*.ts": [
       "tslint --project tsconfig.base.json --fix",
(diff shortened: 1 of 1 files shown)


dhhyi:
Too slow.

PR #498, opened 2021-01-12
fix: user registration mail always sent in english

shauke:
## PR Type

[x] Bugfix

User registration email is always sent in english (en_US) even if the user triggers the registration from a different storefront locale and with a this different locale in the REST call too.
This results from the hardcoded 'en_US' as preferredLanguage in the registration form.

## What Is the New Behavior?

The currently selected locale is used as 'preferredLanguage' when sending the registration form. With this change the registration email is send in the locale the user triggers the registration from.

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/services/user/user.service.ts
+++ b/src/app/core/services/user/user.service.ts
@@ -85,29 +85,36 @@ export class UserService {
       ...body.address,
       mainDivision: body.address.mainDivisionCode,
     };
-    let newCustomer: CreatePrivateCustomerType | CreateBusinessCustomerType;
 
-    if (!body.customer.isBusinessCustomer) {
-      newCustomer = {
-        type: 'PrivateCustomer',
-        ...body.customer,
-        ...body.user,
-        address: customerAddress,
-        credentials: body.credentials,
-      };
-    } else {
-      newCustomer = {
-        type: 'SMBCustomer',
-        ...body.customer,
-        user: { ...body.user },
-        address: customerAddress,
-        credentials: body.credentials,
-      };
-    }
+    let newCustomer$: Observable<CreatePrivateCustomerType | CreateBusinessCustomerType>;
+    newCustomer$ = this.appFacade.currentLocale$.pipe(
+      map(currentLocale =>
+        body.customer.isBusinessCustomer
+          ? {
+              type: 'SMBCustomer',
+              ...body.customer,
+              user: {
+                ...body.user,
+                preferredLanguage: currentLocale.lang ?? 'en_US',
+              },
+              address: customerAddress,
+              credentials: body.credentials,
+            }
+          : {
+              type: 'PrivateCustomer',
+              ...body.customer,
+              ...body.user,
+              address: customerAddress,
+              credentials: body.credentials,
+              preferredLanguage: currentLocale.lang ?? 'en_US',
+            }
+      )
+    );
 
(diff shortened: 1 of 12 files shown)


dhhyi (changes requested):
Hmm, I think, if this field is required by the API and not actually displayed on the registration form, why not just add it before sending in the [UserService.createUser](https://github.com/intershop/intershop-pwa/blob/7e357aca9c63bc85f9591411069316ec3aefd034/src/app/core/services/user/user.service.ts#L89) method. You also have the AppFacade there, you can use `withLatestFrom` in the stream and you can use '??' to set the parameter if it is not yet set.

src/app/core/services/user/user.service.ts
```diff
@@ -25,6 +25,7 @@ import { ApiService, AvailableOptions } from 'ish-core/services/api/api.service'
 interface CreatePrivateCustomerType extends CustomerData {
   address: Address;
   credentials: Credentials;
+  user?: User;
```
shauke, line 28: Any idea how I could get rid of this not correct addition to satisfy this line?
`{ ...newCustomer, user: { ...newCustomer.user, preferredLanguage: currentLocale.lang ?? 'en_US' } }`
Does this change reflect the idea you suggested?

src/app/core/services/user/user.service.ts
```diff
@@ -107,11 +108,16 @@ export class UserService {
...
-          .post<void>(AppFacade.getCustomerRestResource(body.customer.isBusinessCustomer, isAppTypeRest), newCustomer, {
-            captcha: pick(body, ['captcha', 'captchaAction']),
-          })
+          .post<void>(
+            AppFacade.getCustomerRestResource(body.customer.isBusinessCustomer, isAppTypeRest),
+            body.customer.isBusinessCustomer
+              ? { ...newCustomer, user: { ...newCustomer.user, preferredLanguage: currentLocale.lang ?? 'en_US' } }
+              : { ...newCustomer, preferredLanguage: currentLocale.lang ?? 'en_US' },
+            { captcha: pick(body, ['captcha', 'captchaAction']) }
+          )
```
shauke, line 120: Does this change reflect the idea you suggested?

dhhyi: not quite the way I imagined. Nevertheless: let's be realistic: How fancy can we _actually_ become?! I guess, there's an optimum. But forget about that and be *realistic*. Just one method, but at the *right place* (THIS IS IMPORTANT!!!!!!). I'm good with that 🥰 
https://www.youtube.com/watch?v=gLCduDJVksc

shauke: Let's give it one more try with the `newCustomer$` stream. ;)

src/app/core/services/user/user.service.ts
```diff
@@ -85,29 +85,36 @@ export class UserService {
...
-        type: 'SMBCustomer',
-        ...body.customer,
-        user: { ...body.user },
-        address: customerAddress,
-        credentials: body.credentials,
-      };
-    }
+
+    let newCustomer$: Observable<CreatePrivateCustomerType | CreateBusinessCustomerType>;
+    newCustomer$ = this.appFacade.currentLocale$.pipe(
```
dhhyi, line 90: Looks good 👍

dhhyi (approved):

PR #524, opened 2021-02-01
fix: email and login need to be the same

DiverDori:
## PR Type

[x] Bugfix

Updating email is prevented for b2b users, because email and login need to be the same. 

## What Is the New Behavior?
Now the login is set to the same as the email. 

[ ] Yes
[x] No


SGrueber (changes requested):
For security reasons the user has to enter his password if he wants to  change his login. Until now the user could only change his email so the password input was not necessary. See also the change password page in the PWA and the change email page in the RRS.

PR #527, opened 2021-02-02
fix: punchout improvements

shauke:
## PR Type

[x] Bugfix

## What Is the New Behavior?

Add route guard for 'punchout' routes depending on the feature toggle

[x] No

### DIFF
--- a/src/app/pages/account/account-page.module.ts
+++ b/src/app/pages/account/account-page.module.ts
@@ -2,6 +2,7 @@ import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 
 import { AuthorizationToggleGuard } from 'ish-core/authorization-toggle.module';
+import { FeatureToggleGuard } from 'ish-core/feature-toggle.module';
 import { SharedModule } from 'ish-shared/shared.module';
 
 import { AccountOverviewPageModule } from '../account-overview/account-overview-page.module';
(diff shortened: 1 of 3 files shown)


src/app/pages/account/account-page.module.ts
```diff
@@ -41,53 +47,48 @@ const accountPageRoutes: Routes = [
...
+        path: 'order-templates',
+        data: { breadcrumbData: [{ key: 'account.ordertemplates.link' }] },
+        loadChildren: () =>
+          import('../../extensions/order-templates/pages/order-templates-routing.module').then(
+            m => m.OrderTemplatesRoutingModule
+          ),
+      },
+      {
+        path: 'punchout',
+        canActivate: [FeatureToggleGuard, AuthorizationToggleGuard],
```
dhhyi, line 60: this route is now protected twice (once here, once in the routing-module (should be preferred))

dhhyi (changes requested):
the second commit should have prefix "refactor"

src/app/pages/account/account-page.module.ts
shauke: These are two different routes that need to be protected. This FeatureToggleGuard configuration protects the route `/account/punchout` while the configuration in `punchout-routing.module.ts` protects the route `/punchout`.

dhhyi: but the routing module is the same, right?
here its protecting `extensions/punchout/pages/punchout-routing.module` and in that routing module, there also is a feature toggle protection.
Am I misreading something? 🤔

dhhyi: Ah, now I see the reason for my confusion. There are two routing modules in one file. This violates the [rule of one](https://angular.io/guide/styleguide#rule-of-one) where we should be strict, wen it comes to Angular artifacts. (Different story with models)

The `AccountPunchoutRoutingModule` should be renamed and located to the parallel file `punchout-account-routing.module.ts` (which is an allowed pattern, as far as I can see...)

dhhyi (changes requested):

src/app/pages/account/account-page.module.ts
shauke: This is Angular style rule "01-01". We should definitely not mess with this first one. 😉 
Sorry for confusing you by doing such an unthinkable violation. I was thinking about separating these routes but it did not hurt me as much. I hopefully learned my lesson.
The account routing is now in it's own file.

dhhyi (approved):

PR #531, opened 2021-02-03
perf: tune webpack chunk settings

dhhyi:
## PR Type

[x] Build-related changes

Multiple small bundles produced by webpack. 3rd party imports all over the place.

## What Is the New Behavior?

- main bundle
- vendor bundle (should be the same as long as we don't upgrade dependencies)
- common bundle for all lazy loaded code
- budgets adapted

-> fewer files should lead to an overall performance improvement
initial load size still stays the same, though

[ ] Yes
[x] No

## Other Information

https://medium.com/dailyjs/how-did-angular-cli-budgets-save-my-day-and-how-they-can-save-yours-300d534aae7a
https://webpack.js.org/plugins/split-chunks-plugin/

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -58,12 +58,26 @@
               "sourceMap": false,
               "namedChunks": false,
               "extractLicenses": true,
-              "vendorChunk": false,
+              "vendorChunk": true,
               "buildOptimizer": true,
               "budgets": [
                 {
+                  "type": "bundle",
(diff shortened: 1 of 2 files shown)


dhhyi:
Vendor Bundle Hash stays constant:
image

dhhyi:
integrated into #537

PR #536, opened 2021-02-06
feat: multi-channel with multi ICM support

### LINKED ISSUE #420: Support for multiple ICM systems with one PWA

Currently Multi Site support allows for using one PWA for several channels/applications of the same ICM instance. However in a typical setup there are multiple ICM systems e.g. edit and live systems for INT UAT and PRD environments each.

For review and testing of ICM, usually a PWA is needed. Currently one PWA per ICM would be needed, resulting in 6 identical instances of the PWA to accommodate for the above scenario.

It should be possible to use one PWA instance and connect it to several ICM systems, similar to the current Multi Site support.

dhhyi:
## PR Type

[x] Feature

Issue Number: Closes #420

## What Is the New Behavior?

Multi-Channel support is also multi-ICM support 🎉 

[ ] Yes
[x] No

### DIFF
--- a/docs/guides/nginx-startup.md
+++ b/docs/guides/nginx-startup.md
@@ -51,6 +51,12 @@ All other properties are optional:
 - **lang**: The default language as defined in the Angular CLI environment
 - **theme**: The theme used for the channel (format: `<theme-name>(|<icon-color>)?`)
 
+Dynamically directing the PWA to different ICM installations can by done by using:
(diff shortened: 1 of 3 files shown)


jometzner (approved):
I played around with gomplate locally. It suits the need. Even though I'm more a fan of using the tools the image already provides.

Thanks Dan.

PR #540, opened 2021-02-08
feat: use seo metadata provided by CMS REST API for content pages

### LINKED ISSUE #505: SEO: Support Meta Tags for Static Content Pages

The configuration parameters for the CMS Meta Tags (Title, Description, Keywords) of the Static Content Page Component are being ignored by PWA. Instead, the DisplayName of the Page is being used as Title.

Meta Tags (Title, Description, Keywords) of static content pages, when configured via CMS, will be correctly set in the page header.

dhhyi:
## PR Type

[x] Feature

Issue Number: Closes #505

[ ] Yes
[x] No


src/app/core/models/content-pagelet-entry-point/content-pagelet-entry-point.mapper.ts
```diff
@@ -43,6 +44,7 @@ export class ContentPageletEntryPointMapper {
       resourceSetId: data.resourceSetId,
       pageletIDs,
       configurationParameters,
+      seoAttributes: SeoAttributesMapper.fromCMSData(pagelets.length && pagelets[0]),
```
shauke, line 47: It would be good to add the seoAttributes only for content pages and not for content includes.

Suggestion: Add content entrypoint `page` flag if fetched via `/page` REST resource and use this to distinguish if to apply the `SeoAttributesMapper.fromCMSData` or not.

shauke (changes requested):

src/app/core/models/content-pagelet-entry-point/content-pagelet-entry-point.mapper.ts
dhhyi: Hmm, then I'd rather do it outside altogether. I also didn't really like this approach, as it mixes concerns. 😉

shauke (approved):
You are keeping me on my toes to try to understand how it can be done the fancy way. 😉

dhhyi:
> You are keeping me on my toes to try to understand how it can be done the fancy way. 😉

I hope it's also fun for you as well 😉

PR #538, opened 2021-02-07
feat: deploy PWA without SSR

dhhyi:
## PR Type

[x] Feature

## What Is the Current Behavior?

No Dockerfile for deployments without SSR

## What Is the New Behavior?

Dockerfile for Deployment without SSR
Simply applying https://github.com/intershop/intershop-pwa/blob/develop/docs/concepts/deployment-angular.md#simple-browser-side-rendering

## Does this PR Introduce a Breaking Change?

[ ] Yes
[x] No

# Other Information

For this to work with Multi-Channel support, the `configurationMeta` reducer must be active on the client side, which it isn't currently by default:
https://github.com/intershop/intershop-pwa/blob/7b3d937090e00920996b7a0edec5619924dd6c07/src/app/core/store/core/core-store.module.ts#L36

PR #548, opened 2021-02-12
chore: script for checking syncronizations

dhhyi:
## PR Type

[x] CI-related changes

- #538 introduces code duplication in Dockerfile, that should stay in sync, but is not assured so by the CI

## What Is the New Behavior?

- script checks synchronization in CI

[ ] Yes
[x] No


dhhyi:
integrated into #549

PR #572, opened 2021-02-24
build: make it possible to test and review in production mode

dhhyi:
## PR Type

[x] Build-related changes
[x] CI-related changes

With the webpack custom build we introduced a couple of minifications for production, but neither the cypress tests nor the demo servers are deployed with production settings to review and automatically test, if everything also works in production mode.

## What Is the New Behavior?

The GitHub review PWAs are deployed with production settings which also includes CSS minification for visual review.
See here for example:
- http://pwa-gh-review-572-universal-b2c.azurewebsites.net/page/page.helpdesk.pagelet2-Page
- https://intershoppwa.azurewebsites.net/page/page.helpdesk.pagelet2-Page
-> orange boxes are missing because this class is purged for production (thanks to @shauke for this example)

Now also cypress testing is run with production mode (and still ngrx runtime checks and data-testing-ids).

One test (e2e/cypress/integration/specs/checkout/shopping-prices.b2b.e2e-spec.ts) was always failing because fetching the company user returned an empty response from the ICM when it is done in the form `/customers/-/users/-`. When using the customer number in that request everything works. This is supposed to be the new standard, not to ask with `-` anymore, right? 🤔 

While debugging this, because I couldn't reconstruct this problem locally, I also added [cypress-log-to-output](https://www.npmjs.com/package/cypress-log-to-output) plugin to cypress, which now also prints browser logs (in a very verbose fashion) to the CI log output for debugging.

[ ] Yes
[x] No

### DIFF
--- a/.github/workflows/demo-server.yml
+++ b/.github/workflows/demo-server.yml
@@ -33,14 +33,15 @@ jobs:
         uses: elgohr/Publish-Docker-Github-Action@master
         env:
           serviceWorker: false
-          configuration: local
+          configuration: production
           displayVersion: ${{ github.event.after }}
+          testing: true
         with:
           name: universal
           username: ${{ secrets.DOCKER_REGISTRY_USERNAME }}
           password: ${{ secrets.DOCKER_REGISTRY_PASSWORD }}
           registry: ${{ secrets.DOCKER_REGISTRY }}
-          buildargs: serviceWorker,configuration,displayVersion
+          buildargs: serviceWorker,configuration,displayVersion,testing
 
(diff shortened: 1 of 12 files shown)


dhhyi:
The login form also became bigger 🤔 😅 

Before:
image

After:
image

shauke:
> 
> 
> The login form also became bigger 🤔 😅
> 
> Before:
> image
> 
> After:
> image

This was a purgecss safelisting issue. Is already fixed in the develop branch.

shauke (dismissed):

shauke (approved):

PR #583, opened 2021-02-26
feat: detect browser language

### LINKED ISSUE #104: Set language based on browser setting

Users should automatically see the storefront in their preferred language if possible. 

If the users enters the storefront and the browser locale is equal to one that is supported by the storefront, it should be used by default by the storefront. In case the user wants a different language, she can easily switch th the desired locale via language switch.

dhhyi:
## PR Type

[x] Feature

## What Is the New Behaviour?

- Closes #104
- nginx implements simple browser language detection (extracts the first locale from Accept-Language headers) and passes to SSR

[ ] Yes
[x] No

## Other Information

- Language detection has to be implemented in nginx, to assure the correct responses are also cached there.

## Scenarios

- Scenarios, that are handled by this:
  - Europe-wide Shop (shop.com) with all possible languages configured
  - Shop-Channel in Belgium or Switzerland, where multiple languages apply (fr-BE, nl-BE, ...)
- Scenarios not handled:
  - Europe-wide shop with different channels per country. For this a pop-up when reaching the page should suggest switching to the specific shop of the country -> #588

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
       # PROMETHEUS: 1
       # COMPRESSION: 0
       # DEVICE_DETECTION: 0
+      LOCALE_DETECTION: 1
       # MULTI_CHANNEL_SOURCE: env:///ASDF?type=application/yaml
       MULTI_CHANNEL: |
         .+\.net:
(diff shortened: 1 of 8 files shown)


shauke:
The feature of this pull request needs to be reviewed/evaluated together with the suggested changes from #588 (feat: implement channel change feature).

dhhyi:
I split the original PR to only contain the actual language detection. Other refactorings/features are in #658, #657, #659 and #661

wait for integration of #661

PR #594, opened 2021-03-02
fix: map server-config value for null correctly

dhhyi:
## PR Type

[x] Bugfix

Server config settings with value `null` are mapped to an empty object `{ }`.
see:
image
image

## What Is the New Behavior?

`null` is correctly mapped.

[ ] Yes
[x] No

## Other Information

This happens because `typeof null === 'object'` and no special check was in place.


dhhyi:
@shauke please keep commit from #604 as individual commit 😄 . cc @SGrueber

shauke (approved):

PR #603, opened 2021-03-06
refactor: remove redundant productAsVariationProduct from product context

dhhyi:
## PR Type

`ProductContextFacade` provides access to property `product` and `productAsVariationProduct`. The latter which is only set if the product is a `VariationProductView`. This adds additional types which can then be used in components and templates.

It would however be easier if the templates and components could just use all available product properties from the start.

## What Is the New Behavior?

- refined type `ProductView` to include all properties of normal products, master products and variation products.
  - type distinction on view level is mostly hindering than helpful
  - type distinction for basic `Product` models is helpful in mapper and helper.
- removed redundant property `productAsVariationProduct` from `ProductContextFacade`.
- unified `Product` models.

[ ] Yes
[x] No


MaxKless (approved):
looka good, nice type-fu :+1:

PR #611, opened 2021-03-09
fix: allow parentheses in urls

### LINKED ISSUE #610: PWA can't handle "(someval)" in the URL
When the URL contains (someval), the PWA 'crashes' and after the timeout results in a 500 error. In case of the demo website, you'll get the nginx error page.

As far as I can tell, the PWA doesn't create any URL's with the (someval), but that doesn't mean the site should crash. 
In our situation, we have some redirects in place from the old to the new site and those do have this in the URL. (or at least Google thinks so).

dhhyi:
## PR Type

[x] Bugfix

Issue Number: Closes #610 

## What Is the New Behavior?

- custom URL serializer added, that encodes parentheses while processing URLs

[ ] Yes
[x] No

## Other Information

Sadly, the custom serializer has to be provided in `AppModule`. I did not get it working, when providing it in `CoreModule`, where such stuff usually belongs to...

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,9 +1,11 @@
 import { NgModule } from '@angular/core';
 import { BrowserModule, TransferState } from '@angular/platform-browser';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { UrlSerializer } from '@angular/router';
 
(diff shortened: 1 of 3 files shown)


jometzner (approved):

dhhyi:
nice testing @jometzner 😉👍

jometzner:
> nice testing @jometzner

:relaxed: no brainer

PR #612, opened 2021-03-10
chore: run ngcc after build for docker images and dev container

dhhyi:
## PR Type

[x] Build-related changes

- dependencies are ivy compiled when invoking build commands in docker images

## What Is the New Behavior?

- dependencies are ivy compiled directly after install for docker images (faster)

[ ] Yes
[x] No

## Other Information

https://angular.io/guide/ivy#ivy-and-universalapp-shell

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -37,7 +37,7 @@
     "yzhang.markdown-all-in-one"
   ],
   "forwardPorts": [4200],
-  "postStartCommand": "sudo chown node node_modules && npm install",
+  "postStartCommand": "sudo chown node node_modules && npm install && npm run ngcc",
   // Comment out the next line to run as root instead.
   "remoteUser": "node"
 }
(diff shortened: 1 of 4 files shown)


dhhyi:
Turns out, it actually doesn't speed up the build, even takes longer on this CI.
It will however improve build speed when incremental docker builds are done. (like locally)

shauke:
>In version 9, the server builder which is used for App shell and Angular Universal has the bundleDependencies option enabled by default. If you opt-out of bundling dependencies you will need to run the standalone Angular compatibility compiler (ngcc). This is needed because otherwise Node will be unable to resolve the Ivy version of the packages.

Did we "opt-out of bundling dependencies"? I cannot see this in our projects configuration, so there seems to be no need for the `ngcc`? I might want to stick with the defaults in that case, especially if there is not really much to gain.

dhhyi:
> Did we "opt-out of bundling dependencies"? I cannot see this in our projects configuration, so there seems to be no need for the `ngcc`? I might want to stick with the defaults in that case, especially if there is not really much to gain.

No we didn't. And yes, the gain is not really that big, let's stick to the standard process.
Closing this one 😉

PR #614, opened 2021-03-11
refactor: use setLoadingOn and unsetLoadingAndErrorOn where possible

MaxKless:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Some reducers set their loading and error states manually when they don't have to.

## What Is the New Behavior?
Wherever possible and sensible, reducers now use setLoadingOn and unsetLoadingAndErrorOn for repetitive logic.

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/store/customer/user/user.reducer.ts
+++ b/src/app/core/store/customer/user/user.reducer.ts
@@ -5,7 +5,7 @@ import { HttpError } from 'ish-core/models/http-error/http-error.model';
 import { PaymentMethod } from 'ish-core/models/payment-method/payment-method.model';
 import { User } from 'ish-core/models/user/user.model';
 import { loadRolesAndPermissionsFail } from 'ish-core/store/customer/authorization';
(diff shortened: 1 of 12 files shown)


src/app/core/store/customer/user/user.reducer.ts
```diff
@@ -97,6 +97,15 @@ export const userReducer = createReducer(
     deleteUserPaymentInstrumentFail,
     loadRolesAndPermissionsFail
   ),
+  unsetLoadingAndErrorOn(
```
SGrueber, line 100: passwordReminder actions are missing in all methods (setLoadingOn, ...)

SGrueber (changes requested):
There are still occurances left in:
- contact.reducer
- basket.reducer (addPromotionCodeToBasketSuccess) 

if we had a method unsetLoading we could use them in the following reducers. maybe we should introduce this method?
- pages.reducer
- regions.reducer
- product-configuration.reducer
- contact.reducers

SGrueber (approved):

PR #616, opened 2021-03-12
feat: replace any typescript file by configuration

dhhyi:
## PR Type

[x] Feature

Environment specific overrides (#537) can only override HTML templates and component styles

## What Is the New Behavior?

Any typescript file can be overridden by having the extension "<key>.ts".

File overrides can be active for multiple configurations. For example `custom.service.brand1.brand2.ts` will activate for `ng build -c brand1` and also for `ng build -c brand2`.

[ ] Yes
[ ] No

## Other Information

This is useful for projects for overriding functionality without introducing changes in ISH sources.

### DIFF
--- a/docs/guides/customizations.md
+++ b/docs/guides/customizations.md
@@ -81,9 +81,11 @@ Typical hot-spots where copying is a good idea are header-related or product-det
 
 #### Environment Specific Copies
 
-The [customized webpack build](./optimizations.md) supports replacing HTML templates and component SCSS files with an environment suffix before the file extension.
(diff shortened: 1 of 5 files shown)


shauke (changes requested):
Nice feature, but some additional documentation would be nice too. ;)

dhhyi:
> Nice feature, but some additional documentation would be nice too. ;)

Okay 😄

dhhyi:
I took the opportunity to also sneak in an optimization and a test. 👍

shauke (approved):

PR #617, opened 2021-03-13
perf: use lazy properties on product context

dhhyi:
## PR Type

- helper methods on product context for getting promotions and links lazy
- bundle and retail set parts are always fetched

## What Is the New Behavior?

- lazy Observable properties in the ProductContext type

[ ] Yes
[x] No

## Other Information

Not fetching bundle and retail set information for product tiles should decrease the home page rendering time.

### DIFF
--- a/src/app/core/store/shopping/products/products.effects.ts
+++ b/src/app/core/store/shopping/products/products.effects.ts
@@ -18,6 +18,7 @@ import {
   mergeMap,
   switchMap,
   switchMapTo,
+  take,
   throttleTime,
   withLatestFrom,
 } from 'rxjs/operators';
@@ -41,12 +42,13 @@ import {
 
 import {
   loadProduct,
-  loadProductBundlesSuccess,
   loadProductFail,
   loadProductIfNotLoaded,
   loadProductLinks,
   loadProductLinksFail,
   loadProductLinksSuccess,
+  loadProductParts,
+  loadProductPartsSuccess,
   loadProductSuccess,
   loadProductVariationsFail,
   loadProductVariationsIfNotLoaded as loadProductVariationsIfNotLoaded,
@@ -55,11 +57,12 @@ import {
   loadProductsForCategoryFail,
   loadProductsForMaster,
   loadProductsForMasterFail,
-  loadRetailSetSuccess,
 } from './products.actions';
 import {
   getBreadcrumbForProductPage,
+  getProduct,
   getProductEntities,
+  getProductParts,
   getProductVariationSKUs,
   getSelectedProduct,
 } from './products.selectors';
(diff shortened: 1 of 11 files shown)


src/app/core/store/shopping/products/products.effects.ts
```diff
@@ -169,18 +172,44 @@ export class ProductsEffects {
...
+      ofType(loadProductParts),
+      mapToPayloadProperty('sku'),
+      groupBy(identity),
+      mergeMap(group$ =>
+        group$.pipe(
+          this.throttleOnBrowser(),
+          mergeMap(sku =>
+            this.store.pipe(
+              select(getProductParts(sku)),
+              first(),
```
MaxKless, line 186: Why is first() and take(1) used in the same pipe? Doesn't the take(1) make first() irrelevant?

src/app/core/store/shopping/products/products.effects.ts
```diff
@@ -169,18 +172,44 @@ export class ProductsEffects {
...
+              switchMap(() => this.store.pipe(select(getProduct(sku)))),
+              filter(product => ProductHelper.isProductBundle(product) || ProductHelper.isRetailSet(product)),
+              take(1)
+            )
+          ),
+          exhaustMap(product =>
+            ProductHelper.isProductBundle(product)
+              ? this.productsService.getProductBundles(product.sku).pipe(
+                  mergeMap(({ stubs, bundledProducts: parts }) => [
+                    ...stubs.map((stub: Product) => loadProductSuccess({ product: stub })),
```
MaxKless, line 197: is it guaranteed that stubs will always contain values? If not, the [array spread operator might cause problems when used with an undefined value](https://github.com/Microsoft/TypeScript/issues/15027).

src/app/core/store/shopping/products/products.effects.ts
```diff
@@ -169,18 +172,44 @@ export class ProductsEffects {
...
+                    loadProductPartsSuccess({ sku: product.sku, parts }),
+                  ]),
+                  mapErrorToAction(loadProductFail, { sku: product.sku })
+                )
+              : this.productsService.getRetailSetParts(product.sku).pipe(
+                  mergeMap(stubs => [
+                    ...stubs.map((stub: Product) => loadProductSuccess({ product: stub })),
```
MaxKless, line 204: ^ same here

MaxKless (commented):
Looks good to me, very nice typing :D Just some questions about the long pipe and I'll approve.

src/app/core/store/shopping/products/products.effects.ts
dhhyi: Not in this case. Every store select must be aborted after the first result, otherwise the effect keeps on firing when the result is set. In this case though, the following filter would also implicitly prevent that.

src/app/core/store/shopping/products/products.effects.ts
dhhyi: It is defined because the mapper will at least yield an empty array. It would be a ICM configuration issue otherwise, if no parts are available. So an unexpected error that is not swallowed is also good. Don't you think? 😅

src/app/core/store/shopping/products/products.effects.ts
MaxKless: Good to have it explicit and it's nice to have this comment for future reference in case someone wonders :D

MaxKless (approved):

PR #626, opened 2021-03-19
feat: display only available products in product links carousels and lists

carlosDjangoo:
## PR Type

[x] Feature

The product links carousels/lists (cross-sellings, up-sellings. accessories, replacements, etc.) have no option to filter products out that are not available.

## What Is the New Behavior?

The product links carousel and product links list now have an input parameter `displayOnlyAvailableProducts` to configure if you want to display only available products. Via the parameter it can be controlled which lists should only present sell-able products (e.g. up-selling, cross-selling) and which lists should list products even if they are not available right now (e.g. spare parts).

[x] No

### DIFF
--- a/src/app/pages/product/product-links-carousel/product-links-carousel.component.html
+++ b/src/app/pages/product/product-links-carousel/product-links-carousel.component.html
@@ -1,9 +1,9 @@
-<div *ngIf="links.products.length" class="product-list-container">
+<div *ngIf="productSKUs.length" class="product-list-container">
   <h2>{{ productLinkTitle }}</h2>
 
   <div class="product-list">
     <swiper [config]="swiperConfig">
-      <div *ngFor="let sku of links.products" class="swiper-slide">
+      <div *ngFor="let sku of productSKUs" class="swiper-slide">
         <ng-template swiperSlide>
           <ish-product-item ishProductContext [sku]="sku"></ish-product-item>
         </ng-template>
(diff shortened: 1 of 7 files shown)


dhhyi:
@carlosDjangoo 
The `ShoppingFacade.products$` doesn't exist anymore in the current version. Consider upgrading, product contexts are awesome 😉 
The `product$` stream from `ShoppingFacade` was the right choice. I don't think we currently have a REST call to fetch multiple products individually (only for categories or master products). This would be a nice feature though, which could be used in `products.effects`.

I supplied a fixup commit, that purely focuses on RxJS streams, which makes helper methods and `destroy$` obsolete.

I think the setting to `true` in `src/app/pages/product/product-links/product-links.component.html` should be considered a project customization and not be part of this PR, but that is just an opinion. 😄

carlosDjangoo:
> @carlosDjangoo
> The `ShoppingFacade.products$` doesn't exist anymore in the current version. Consider upgrading, product contexts are awesome 😉
> The `product$` stream from `ShoppingFacade` was the right choice. I don't think we currently have a REST call to fetch multiple products individually (only for categories or master products). This would be a nice feature though, which could be used in `products.effects`.
> 
> I supplied a fixup commit, that purely focuses on RxJS streams, which makes helper methods and `destroy$` obsolete.
> 
> I think the setting to `true` in `src/app/pages/product/product-links/product-links.component.html` should be considered a project customization and not be part of this PR, but that is just an opinion. 😄

@dhhyi  Thanks this is a good idea with rxjs 😉 The setting 'true' in product-links.component.html was only to show you an example, this is part of project you are right. An individually product rest call  would be a nice feature, because some customers want to have individual product lists and suggestions. And call every product individually can be run in a performence issue then.

SGrueber:
The configuration parameter can be useful in projects, that's why we decided to leave the code as it is. We added code for the product links list, additionally.

shauke (approved):

shauke (approved):

PR #636, opened 2021-03-25
chore: sort localizations in pre-commit

dhhyi:
## PR Type

[x] Build-related changes

- localization clean is tested on CI

## What Is the New Behavior?

- localization cleanup in precommit

[ ] Yes
[x] No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19506,6 +19506,31 @@
         "socks": "^2.3.3"
       }
     },
+    "sort-json": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/sort-json/-/sort-json-2.0.0.tgz",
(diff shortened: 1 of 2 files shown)


shauke (commented):
I would rather not merge this change as it might remove localizations unintended. It is probably better to have the CI fail and intentionally clean up the not needed localizations then silently removing them. We might need to discuss this in the development team/community?

dhhyi:
@shauke Good point. I was suggesting this change for the sorting. Maybe I should add a --precommit flag, that fails if a key would be removed and then asks the programmer to run it manually again?

dhhyi:
> Maybe I should add a --precommit flag, that fails if a key would be removed and then asks the programmer to run it manually again?

Better idea: use a tool dedicated to sorting json 😉

shauke (approved):

PR #644, opened 2021-04-01
fix: auth0 silentRefresh requires offline_access scope

franciscomdrito:
see: https://manfredsteyer.github.io/angular-oauth2-oidc/docs/additional-documentation/refreshing-a-token.html

image

<!--

## PR Type
[X] Bugfix

The authentication flow uses an automatic silent refresh, which is responsible to refresh the id and auth token before it expires. Currently this refresh token is failing: 
image

And according to the documentation, the offline_access scope is required to trigger the silent refresh.

image

## What Is the New Behavior?
Adding the offline_access scope will properly refresh the tokens before they expire.

[ ] Yes
[X] No

### DIFF
--- a/src/app/core/identity-provider/auth0.identity-provider.ts
+++ b/src/app/core/identity-provider/auth0.identity-provider.ts
@@ -66,7 +66,7 @@ export class Auth0IdentityProvider implements IdentityProvider {
 
       // Scopes ("rights") the Angular application wants get delegated
       // https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
-      scope: 'openid email profile',
+      scope: 'openid email profile offline_access',
 
       // Using Authorization Code Flow
       // (PKCE is activated by default for authorization code flow)


dhhyi (approved):

jometzner:
Weird thing is that we currently only use the ID token to authenticate against REST resources. I thought that's the job of the Access token. Am I overseeing something?

dhhyi:
> Weird thing is that we currently only use the ID token to authenticate against REST resources. I thought that's the job of the Access token. Am I missing something?

You are right. The [angular-oauth2-oidc library](https://manfredsteyer.github.io/angular-oauth2-oidc/docs/additional-documentation/working-with-httpinterceptors.html) also provides an automatism to send access_token, but the ICM has different requirements.

jometzner:
I double checked this issue on my demo setup with auth0.com acting as OIDC backend.

1. Changed ID Token expiration time to 5 min
2. Logged in
3. Waited 5 minutes
4. Saw the depicted behavior
My Account | Intershop PWA - Vivaldi_031
Selection_032

jometzner (approved):

PR #655, opened 2021-04-17
feat: load translations with webpack

dhhyi:
## PR Type

[x] Feature

`ngx-translate` translation files are loaded with [TranslateHttpLoader](https://github.com/ngx-translate/http-loader#1-setup-the-translatemodule-to-use-the-translatehttploader) on the browser side and a special translate loader on the server side. Due to this handling, the translation files are not automatically output-hashed via Angular CLI. This can lead to misbehavior, when new keys are added but the browser still holds the old translation file in the browser cache.

## What Is the New Behavior?

- Translation files are consistently loaded via webpack and automatically output-hashed
- ~~The old mechanism of `registerLocaleData` is exchanged with the [new one](https://angular.io/guide/i18n#import-global-variants-of-the-locale-data)~~ Doesn't work with FF
- Translation file ~~and Angular locale data~~ is put into its own bundle and pulled when needed.

[ ] Yes
[x] No

### DIFF
--- a/docs/concepts/configuration.md
+++ b/docs/concepts/configuration.md
@@ -219,21 +219,27 @@ To dynamically set the default locale, use the URL parameter `lang` when rewriti
 
 ## Extend Locales
 
-To add other languages except English, German or French, you have to create a new json-mapping-file with all translations, e.g., _./src/assets/i18n/nl_NL.json_).
-Add the locale in the file _./src/environment/environments.ts_.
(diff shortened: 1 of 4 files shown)


MaxKless:
Hi Danilo, thanks for the PR.
We had some issues with this PR but I see you saw some webpack issues yourself and fixed them. As in #661, we built our own loader during the localization rework (#695). We are not generally opposed to this change. If you want to, you can rebase on our latest release and reimplement these changes. If you decide against it you can close this PR and we will track this enhancement ourselves.
Thanks again!

dhhyi:
@MaxKless I rebased and put in some refactoring. Please test if everything still works properly.

MaxKless (approved):

PR #659, opened 2021-04-18
feat: improve detecting missing translations

dhhyi:
## PR Type

[x] Feature

- Missing translations are silently replaced with the ones from the ngx-translate default lang or just the translation key is inserted
- The default lang is always pulled for client applications

## What Is the New Behavior?

- Missing translations trigger an error in the console, so that they can be detected (SSR logs, Sentry, Browser console)
- Pulling default fallbacks can be customized
- final SSR locale is transferred to the client to be used for initialization

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/internationalization.module.ts
+++ b/src/app/core/internationalization.module.ts
@@ -1,16 +1,17 @@
-import { isPlatformBrowser, isPlatformServer, registerLocaleData } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
+import { registerLocaleData } from '@angular/common';
 import localeDe from '@angular/common/locales/de';
 import localeFr from '@angular/common/locales/fr';
-import { Inject, Injectable, LOCALE_ID, NgModule, PLATFORM_ID } from '@angular/core';
+import { Inject, Injectable, LOCALE_ID, NgModule } from '@angular/core';
 import { TransferState } from '@angular/platform-browser';
+import { Actions, ROOT_EFFECTS_INIT, ofType } from '@ngrx/effects';
 import { Store, select } from '@ngrx/store';
 import { MissingTranslationHandler, TranslateLoader, TranslateModule, TranslateService } from '@ngx-translate/core';
-import { from, of } from 'rxjs';
-import { catchError, map, switchMap, take, tap, withLatestFrom } from 'rxjs/operators';
+import { ReplaySubject, combineLatest, defer, from, of } from 'rxjs';
+import { catchError, first, map, switchMap, take, tap } from 'rxjs/operators';
 
-import { SSR_LOCALE, SSR_TRANSLATIONS } from './configurations/state-keys';
-import { getRestEndpoint } from './store/core/configuration';
+import { NGRX_STATE_SK, STATE_ACTION_TYPE } from './configurations/ngrx-state-transfer';
+import { SSR_LOCALE } from './configurations/state-keys';
+import { getServerTranslations, loadServerTranslations } from './store/core/configuration';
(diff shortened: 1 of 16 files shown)


dhhyi:
wait for integration of #655

MaxKless:
Hi Danilo, thanks for the PR.
During our localization rework (#695) we adopted some of your SSR-specific code. Thank you very much for your contribution. 
The missing translation handling is definitely a nice-to-have but not a priority for us right now. You're very welcome to rebase and integrate your changes with the current internationalization module and I will look at it ASAP! Otherwise, we will close this PR and track the enhancement ourselves.

dhhyi:
@MaxKless I went for the rebasing. Please have a look 😄

src/app/core/internationalization.module.ts
```diff
@@ -83,12 +81,46 @@ class ICMTranslateLoader implements TranslateLoader {
...
+    private translateLoader: TranslateLoader,
+    private errorHandler: ErrorHandler,
+    @Inject(FALLBACK_LANG) private fallback: string
+  ) {}
+
+  handle(params: MissingTranslationHandlerParams) {
+    if (params.interpolateParams || /^\w+(\.[\w-]+)+$/.test(params.key)) {
+      this.errorHandler.handleError(`missing translation in ${params.translateService.currentLang}: ${params.key}`);
+      if (params.translateService.currentLang !== this.fallback) {
+        return this.translateLoader.getTranslation(this.fallback).pipe(
```
MaxKless, line 99: While I see how it's nice to consolidate the fallback handling into the MissingTranslationsLoader, the current setup triggers a `getTranslation()` call for each missing translation, resulting in many repetitive calls. It should be possible to cache the fallback translations to avoid this.

Also, the API provides key-specific translation functionality:
image
So in case of a missing translation key (maybe from an error message) it would be nice to check the API for that specific key in addition to querying the fallback translation.

src/app/core/store/core/configuration/configuration.effects.ts
```diff
@@ -53,12 +52,10 @@ export class ConfigurationEffects {
         select(getCurrentLocale),
         mapToProperty('lang'),
         distinctUntilChanged(),
-        // https://github.com/ngx-translate/core/issues/1030
-        debounceTime(0),
         whenTruthy(),
         switchMap(lang => languageChanged$.pipe(mapTo(lang), take(1)))
       )
-      .subscribe((lang: string) => {
+      .subscribe(lang => {
```
MaxKless, line 58: Where did the typing go? Is there a downside to it?

MaxKless (changes requested):
Thanks for the work but I think the fallback handling can still be improved.

src/app/core/store/core/configuration/configuration.effects.ts
dhhyi: It's no longer necessary because of the `mapTo` in the previous line.

src/app/core/internationalization.module.ts
dhhyi: ✔️

dhhyi: I would suggest limiting the individual lookup to the client and only to error messages. Otherwise too many calls would be triggered.

Still haven't got it working perfectly, I'll have another look tomorrow 😄

dhhyi: I refactored the feature to use the NgRx store and now it seems to be working 😄

MaxKless (approved):
All around great changes!!

PR #661, opened 2021-04-18
feat: respect icm language settings

dhhyi:
## PR Type

[x] Feature

The PWA ignores locale settings from the ICM, that are available in the `configurations` REST response.

## What Is the New Behavior?

If retrieved, the PWA uses the ICM locale settings to set a default locale and filter available locales based on channel settings.

[ ] Yes
[x] No

## Other Information

To have access t the configurations call results in the configurations store, the server-config store has to be moved from general to core.


dhhyi:
wait for integration of #659

MaxKless:
Hi Danilo, thanks for the PR! When you submitted this, we were in the middle of our own localization rework. You wrote some nice code, though and we used parts of it in #695 .
Thanks for your work but since this is fixed I'm closing the PR.

PR #674, opened 2021-04-30
fix: in case of multiple buckets display message on checkout review page

SGrueber:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

1. If the basket contains items that cannot be shipped together, but it is complete and validated successfully, the user is navigated from the cart page to the review page (in case of basket acceleration), but the review page is corrupt.
2. If the user submits a basket but the order could not be created (e.g. because a payment is not authorized), an error is displayed on checkout review page. If the user navigates to another page, e.g. the payment page and fixes the problem and he navigates to the review page again, the order creation error is displayed again.

Issue Number: Closes #

## What Is the New Behavior?
1. If items cannot be shipped together an error message is shown on checkout review page and the submit order button is disabled.
2. After navigating from the checkout review page order creation errors are reseted and they are not shown again, if the user enters the checkout review page.

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/store/customer/basket/basket.effects.spec.ts
+++ b/src/app/core/store/customer/basket/basket.effects.spec.ts
@@ -12,6 +12,7 @@ import { Basket } from 'ish-core/models/basket/basket.model';
 import { BasketService } from 'ish-core/services/basket/basket.service';
 import { CoreStoreModule } from 'ish-core/store/core/core-store.module';
 import { CustomerStoreModule } from 'ish-core/store/customer/customer-store.module';
+import { resetOrderErrors } from 'ish-core/store/customer/orders';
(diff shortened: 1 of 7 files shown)


shauke:
Navigating from a checkout review page with a message regarding the shipping issues

image

back to the checkout shipping page results in a shipping methods list without pricing and details

image

I am also not certain that the error message "There is no shipping method available that allows these products to be shipped together. Please adjust the shopping cart and place several separate orders." is correct since selecting any shipping method would solve the checkout problem.

shauke (approved):
The additionally mentioned issue will be handled separately.

PR #630, opened 2021-03-20
feat: set deploy URL as runtime property

dhhyi:
## PR Type

[x] Feature

## What Is the Current Behavior?

Deploy URL in links to static resources is sometimes used relatively and sometimes absolute.

Issue Number: Closes #624

## What Is the New Behavior?

The deploy URL (which can be set for default Angular projects with [`ng build --deploy-url`](https://angular.io/cli/build)) can be set via environment variable `DEPLOY_URL` to the SSR process.

If not set, the default deployment URL `/` is applied, which ensures, all static assets are correctly loaded from the deployment root. 

This PR also makes it possible to use deeply nested `baseHref`s with a Multi-Channel context path setup.

This PR depends on #655 to make the Angular application independent on the `DEPLOY_URL`.

## Does this PR Introduce a Breaking Change?

[ ] Yes
[x] No

## TODO

- [x] Documentation -> #646

PR #676, opened 2021-05-01
feat: easily override default config for build

dhhyi:
## PR Type

[x] Feature
[x] Other: Customization enhancement

If a customer project uses a different default build configuration than `production`, the new configuration has to be inserted into quite a lot of scripts in `package.json`.

## What Is the New Behavior?

The default configuration can be overridden in `package.json` or in local npm settings and automatically applies to all scripts.

[ ] Yes
[x] No


dhhyi:
wait for integration of #630

dhhyi:
Will be integrated into #775

PR #677, opened 2021-05-03
feat: basic auth support for nginx

dhhyi:
## PR Type

[x] Feature

For Intershop CaaS environments, UAT and INT deployments are configured to use basic auth, so that the shop is not publicly visible or indexed by search engines.
This setup however is performed on the Kubernetes nginx ingress, where it doesn't differentiate between page responses and retrieval of static assets. Some of the assets can not be retrieved by the PWA without setting up cross-origin use-credentials, and even then (with multiple basic auth dialogs popping up), fonts are not properly retrieved.

## What Is the New Behavior?

The PWA nginx container can be set up using basic auth: https://github.com/intershop/intershop-pwa/blob/feat/nginx-basic-auth/docs/guides/nginx-startup.md#basic-auth
Whitelisting IPs is also supported.
The nginx only guards page responses and makes static assets freely available.

[ ] Yes
[x] No

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,10 @@ services:
       # COMPRESSION: 0
       # DEVICE_DETECTION: 0
       # MULTI_CHANNEL_SOURCE: env:///ASDF?type=application/yaml
+      # BASIC_AUTH: 'developer:!InterShop00!'
+      # BASIC_AUTH_IP_WHITELIST: |
+      #   # - 172.22.0.1
+      #   - 1.2.3.4
       MULTI_CHANNEL: |
         .+:
           - baseHref: /en
@@ -39,6 +43,7 @@ services:
           - baseHref: /de
             channel: default
             lang: de_DE
+            protected: false
           - baseHref: /fr
             channel: default
             lang: fr_FR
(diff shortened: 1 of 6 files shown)


jometzner:
We have to discuss this proposal with our operations team.

psemdel:
Hi, it is Maxime, so we discussed with Evident. Danilo will adapt this PR to make it not anymore global but on channel level. With that it would cover the problem of IP restriction at channel level. Some documentation of the concept would be needed then, to make it clear to all future users.

dhhyi:
Hi @psemdel,
Didn't we conclude at leaving this on global level? And specific projects can then adapt this for channel-level? I think this should be the default use case.
I can however have another look at maybe implementing both variants. 😉

psemdel:
Sorry, I may have misunderstood a point. Do as agreed.

dhhyi:
@psemdel I added an opt-out on channel level.
So if only one of three channels should be publicly accessible, then we can define basic auth globally, but again deactivate it in the multi-channel configuration by setting `protected: false` on the channel. (see [documentation](https://github.com/intershop/intershop-pwa/blob/feat/nginx-basic-auth/docs/guides/nginx-startup.md#basic-auth))
I think this should cover our use case 😉

jometzner (dismissed):
All sunshine and happiness

jometzner (approved):

PR #683, opened 2021-05-07
fix: set punchout cookies with "SameSite=None; Secure" to work in iframes

shauke:
## PR Type
[x] Bugfix

A PWA punchout session does not work within an iframe that is often used by punchout systems, e.g. https://punchoutcommerce.com/tools/cxml-punchout-tester. This is due to necessary cookies (e.g. for the return url) not being set properly in such combinations.

## What Is the New Behavior?
The punchout cookies are set with "SameSite=None; Secure" that allows for the required cookies being set in iframe environments too.

[x] No

### DIFF
--- a/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
+++ b/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
@@ -76,9 +76,18 @@ export class PunchoutPageGuard implements CanActivate {
             return this.punchoutService.getCxmlPunchoutSession(route.queryParamMap.get('sid')).pipe(
               // persist cXML session information (sid, returnURL, basketId) in cookies for later basket transfer
               tap(data => {
-                this.cookiesService.put('punchout_SID', route.queryParamMap.get('sid'), { sameSite: 'Strict' });
-                this.cookiesService.put('punchout_ReturnURL', data.returnURL, { sameSite: 'Strict' });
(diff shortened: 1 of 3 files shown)


src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
```diff
@@ -76,9 +76,18 @@ export class PunchoutPageGuard implements CanActivate {
             return this.punchoutService.getCxmlPunchoutSession(route.queryParamMap.get('sid')).pipe(
               // persist cXML session information (sid, returnURL, basketId) in cookies for later basket transfer
               tap(data => {
-                this.cookiesService.put('punchout_SID', route.queryParamMap.get('sid'), { sameSite: 'Strict' });
-                this.cookiesService.put('punchout_ReturnURL', data.returnURL, { sameSite: 'Strict' });
-                this.cookiesService.put('punchout_BasketID', data.basketId, { sameSite: 'Strict' });
+                this.cookiesService.put('punchout_SID', route.queryParamMap.get('sid'), {
+                  sameSite: 'None',
+                  secure: (isPlatformBrowser(this.platformId) && location.protocol === 'https:') || false,
```
SGrueber, line 81: secure value should be determined only once, either in the guard or in the cookie service

SGrueber (changes requested):
The CookiesOptions interface of the CookiesService should be exported so that it can be used in other source files, too.

SGrueber (approved):

PR #695, opened 2021-05-12
feat : load translations from icm

MaxKless:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently, it is only possible to change translations by updating one of the translation files and re-deploying the PWA.

Issue Number: Closes #

With this PR, it will be possible to update translations in the ICM or somewhere remote and load them into the PWA without redeploying.

## ToDo
- [x] refine process
- [x] Update the loader for localization api
- [x] fix xliff conversion script for object
- [x] ensure SSR functionality
- [x] fix all e2e tests
- [x] docs
 

[ ] Yes
[ ] No


jometzner:
Cypres test is failing due to issue with ICM localization resource (IS-32998).

src/app/core/internationalization.module.ts
```diff
@@ -2,12 +2,95 @@ import { registerLocaleData } from '@angular/common';
...
+}
+
+export function filterAndTransformKeys(translations: Record<string, string>): Translations {
+  const filtered: Translations = {};
+  for (const key in translations) {
+    if (key.startsWith('pwa-')) {
+      const value = translations[key];
+      try {
+        const parsed = JSON.parse(value);
+        filtered[key.replace('pwa-', '')] = parsed;
```
jometzner, line 27: I'd rather use a regex for this replace operation. You never know if that substring is gonna be used at other parts of the key.

marschmidt89 (approved):
documentation reviewed.

src/app/core/internationalization.module.ts
```diff
@@ -16,16 +100,25 @@ export function translateFactory(http: HttpClient) {
...
     registerLocaleData(localeDe);
     registerLocaleData(localeFr);
 
-    translateService.setDefaultLang(lang.replace(/\-/, '_'));
+    let defaultLang = angularDefaultLocale.replace(/\-/, '_');
+    if (transferState.hasKey(SSR_LOCALE)) {
+      defaultLang = transferState.get(SSR_LOCALE, defaultLang);
+    }
+    translateService.setDefaultLang(defaultLang);
+    translateService.use(defaultLang);
```
jometzner, line 122: Why is it necessary to directly call `use` with the default language? This has introduced a race condition in SSR.

jometzner: Well, not quite. But it actually has introduced non deterministic behavior in connection with `configuration.effects.ts`. We need to stick with this call here otherwise we get issue with initing the app.

PR #702, opened 2021-05-17
fix: track item changes as local state in checkout facade

### LINKED ISSUE #668: bug: mini cart opens if the user navigates from checkout to the home page
**Actual Behavior**
The mini cart opens if the user navigates from checkout to the home page and there are items in cart.

**Expected Behavior**
 The mini cart is shown but it doesn't open.

1. Login and add some products to cart
2. Go to cart page and start checkout
3. Navigate to the home page by clicking the logo
4. => mini basket is shown and opens, although no item has been added to cart

dhhyi:
## PR Type

[x] Bugfix

Issue Number: Closes #668

## What Is the New Behavior?

fixed 😁👍 

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/facades/checkout.facade.ts
+++ b/src/app/core/facades/checkout.facade.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Store, createSelector, select } from '@ngrx/store';
-import { merge } from 'rxjs';
-import { debounceTime, distinctUntilChanged, map, shareReplay, switchMap, take, tap } from 'rxjs/operators';
+import { Subject, merge } from 'rxjs';
+import { debounceTime, map, sample, switchMap, take, tap } from 'rxjs/operators';
 
 import { Address } from 'ish-core/models/address/address.model';
 import { Attribute } from 'ish-core/models/attribute/attribute.model';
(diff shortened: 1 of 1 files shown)


src/app/core/facades/checkout.facade.ts
```diff
@@ -68,13 +78,7 @@ export class CheckoutFacade {
   // BASKET
 
   basket$ = this.store.pipe(select(getCurrentBasket));
-  basketChange$ = this.store.pipe(
-    select(getBasketLastTimeProductAdded),
-    whenTruthy(),
-    distinctUntilChanged(),
-    switchMap(() => this.basketLoading$.pipe(debounceTime(500), whenFalsy())),
-    shareReplay(1)
```
dhhyi, line 76: `shareReplay` is not a solution here, because the header is re-initialized via `AppFacade.headerType$` when leaving the checkout. We need a "global" `Observable` that only emits when new changes happen. Using only NgRx is also not a good solution as all subscriptions emit the last value and working with `skip(1)` on subscribe is masking quite a lot (also not working btw). The only probable solution might really be local state in the facade...

cc @shauke @Eisie96

Eisie96 (approved):

PR #699, opened 2021-05-15
deps: Angular 12 update

jometzner:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[X] Refactoring (no functional changes, no API changes)
[X] Build-related changes
[ ] CI-related changes
[ ] Documentation content changes
[ ] Application / infrastructure changes
[ ] Other: 

## What Is the Current Behavior?

Angular 12 is not supported

Issue Number: Closes #

## What Is the New Behavior?

- Updated all angular dependencies to Angular 12
- Refactor custom webpack script
- Update ng-bootstrap

## Does this PR Introduce a Breaking Change?

[ ] Yes
[ ] No

PR #718, opened 2021-06-03
deps: upgrade to Angular 12 with Webpack 5

dhhyi:
## PR Type

[x] Code style update (formatting, local variables)
[x] Build-related changes

Angular 11 with Webpack 4

## What Is the New Behavior?

Angular 12 with Webpack 5

[x] Yes
[ ] No

## Other Information

- bootstrap pinned to version 4
- some jest dependencies (jest@27, jest-preset-angular@9) fail tests on upgrade -- maybe next time 😅 
- prettier upgrade with formatting change as extra commit << please do rebase merge

replaces #699


jometzner:
Looks good. I wasn't that far away with my custom.webpack.ts

angular.json
```diff
@@ -54,6 +54,7 @@
               "namedChunks": false,
               "extractLicenses": true,
               "vendorChunk": true,
+              "commonChunk": false,
```
jometzner, line 57: Why are you disabling the common bundle?

jometzner: Speed up loading time? Only load what is really needed etc.?

dhhyi: it's not disabled, 'technically'. This is just preventing that Angular inserts a default configuration. The webpack.custom still pushes artifacts into the common bundle and it is being created.

jometzner: True that. I was wondering why `ng build --stats-json` still spits out common.js.

jometzner (approved):

PR #738, opened 2021-06-18
feat: add switch for currency selection

dhhyi:
## PR Type

[x] Feature

- Currencies cannot be switched independent of locales

## What Is the New Behavior?

- introduced a currency switch that switches between currencies
  - only available if the ICM channel supports multiple currencies

[ ] Yes
[x] No

## Other Information

- we need this in a project context for choosing USD and CAD for a Canadian channel.


dhhyi:
@shauke Can you have a look? Styling is probably something that you still want to change 😉

MaxKless:
Hi Danilo, thank you for this great PR. It includes a lot of improvement we would like to include in the PWA. There are just two open questions:
- There is no REST request to reload the basket in a new currency - to recalculate prices you have to add/remove something. This will probably cause problems when you switch currency. Have you thought about this issue or is this something that you already have a workaround for?
- With the current setup, it is not possible to configure the currency via a multi-site configuration. Persisting the currency in the URL is not possible, for example when sharing a link with a friend the currency will always reset to default. Is this something you have considered and do you think this would be a problem for your project?

MaxKless:
Thanks for the response @dhhyi :) 
We still need to shine a light behind some uncertainties before we can merge this. Especially the basket recalculation (which is not triggered via the get request) needs to be solved but we're working on that with you. 
This is the things we still have to do and a plan of how to proceed:

- [ ] check UX/VD of the switch -> request review of @M-Behr  @iwiederhold 
- [ ] change ICM  demo data to include multiple currencies (so we can test easily) -> @shauke 
- [ ] figure out how to recalculate basket with the new currency (PATCH request) -> wait for @rowiish to help us understand
- [ ] integrate multi-site configurability -> Review of #630 by @jometzner , code by @dhhyi

dhhyi:
> Here are the things we still have to do and a plan of how to proceed:

Wow, this is blowing up quite fast. May I propose we partition this PR into the refactoring on the PWA side (that enables us to work with our customization) and the integration of the language switch component into the PWA (that enables this feature for the PWA and contains all further tasks?

That would help us @MaxKless

dhhyi:
@MaxKless I stripped away all of the required refactoring to #746, so that this PR only adds the visible part to the Intershop PWA.

rowiish:
Hello @dhhyi, Hello @MaxKless,
> > * There is no REST request to reload the basket in a new currency
> 
> Well, if I add `cur=EUR` to the REST request to the `GET /baskets/current` I would expect that the values returned are all in Euros. This might not be the way the basket API is currently working, though

A `GET` request is not allowed to change anything and that may happen when a recalculation is done (e.g. different promotion applied). That's why you need `PATCH /baskets/current`. Unfortunately, the patch request seems to be broken for the currency change on server side. I'll check that.

In order to trigger a recalculation you have to do a
```
PATCH /baskets/current

{
  "calculated": true
}
```

M-Behr (changes requested):
You are using the correct pattern for the currency switch but unfortunately I found some issues:

- [ ] Replace icon: We would like to use the "money-bill" icon for the currency.
- [ ] Reduce space between language switch and currency switch. See space between compare and wish list as reference.
- [ ] Enlarge the dropdown menu width: US Dollar - no line breaks. The box width should be flexible.
- [ ] Currently the header has not enough space for all the elements in the tablet view. I would recommend to display only the first name in the header. The full name is no important data (the user knows his name). It is more about addressing him personally. The first name is enough for that purpose. Alternatively, we need to remove the words "Compare" & "Wish lists" and display only the icons. But only icons could be misleading.
pwa_header_tablet

Some functional issues:
- [ ] If an item is added to the basket and user changes the currency, the currency in the basket is not updated until you do something with the basket (e.g. add product, change quantity etc.)
- [ ] Currency display is not changed in My Account.
- [ ] Order Approval does not work when you define the budget in $ and buy something in €. The order spent limit is ignored and the order is not displayed in "Already Spent".
PWA_currency_issue
I suggest that the account admin can define a budget per currency (for example: 400USD and 500€) and the display is changed depending on the selected currency. Alternatively, we could add an automatic exchange rate, or we could just change the currency display (but 100USD does not equal 100€) so that at least the approval process works.

rowiish:
>     If an item is added to the basket and user changes the currency, the currency in the basket is not updated until you do something with the basket (e.g. add product, change quantity etc.)

The server side of this functional issue is addressed in AB#46035 and currently in progress.

rowiish:
The ICM issue is fixed with 7.10.34.1. The updated "Concept" document will be become public after the next publish run.

shauke:
Abandoned feature since it was not generally requested/required.
We currently do not intend to include this functionality.

PR #745, opened 2021-06-24
feat: single quickorder - direct add product in cart

### LINKED ISSUE #694: Direct product adding on cart missing

On the PWA version the feature that allows customer to add directly a product on cart is missing

This is a feature present on the classic version. On the cart, a section allows customer to fill product sku and quantity to add one product at his cart.
missing_direct_adding

An alternative is passing by the quick order feature.

Is there something on PWA roadmap, because we need this for our project.

cbratsch:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Issue Number: Closes #694 

## What Is the New Behavior?

Direct order is available on cart page, when quickorder feature is toggled.

[ ] Yes
[x] No


iwiederhold (commented):
Hi, there are some old VD and IAD issues pending.
It would be nice, if these bugs can be fixed in addition.
THX.

M-Behr (changes requested):
It works very well. I found some (smaller) issues:

- [x] Rename error message: "Please enter a valid product number!" => "The product ID 1234 does not exist." => It should be equal with error messages in quick order form. Maybe you can use the same key and we adapt the message for both places.
- [x] "Continue Shopping" link is not part of this section and should be displayed in an extra row. It is something like "Back to Homepage" and it is the last link in this content area. See detail pages in My Account as reference, e.g. Order detail page.
- [x] The "check" icon for the input field overlaps the arrows. Perhaps you can move it a little to the left. Alternatively, we can display the custom input without "Check" icon.
- [x] Please remove the 15px padding between the input fields so that they are closer together.
- [ ] Please move the "Add to Cart" button to the right until end of table. Then we can make the quantity field larger.
- [x] The "Add to Cart" button enlarges when I enter an invalid product ID. Possibly it has something to do with the disabled state.
- [x] The "Add to Cart" button needs some time to switch to disabled state. You can click "Add to Cart" in the meantime. An error message is displayed and the inputs are removed. I would prefer that the entered inputs are displayed in case of an error message. See promotion code behavior as example => Enter an invalid code and click "Apply". The code is still visible in the input field.
- [ ] After successful adding to the cart, please show button in initial state (disabled) to avoid error messages.
- [ ] Some spacing is missing in the responsive layout.

Screenshot - Desktop:
2021-06-28 17_12_08-Your Shopping Cart _ Intershop PWA

Screenshot - Responsive:
2021-06-28 17_07_27-Your Shopping Cart _ Intershop PWA

cbratsch:
> * [ ]  The "check" icon for the input field overlaps the arrows. Perhaps you can move it a little to the left.


This seems to be the first regular number input field we have. So I have different proposals for the check icon.
- move it simply a bit to the left could looks nasty with large numbers 
image
image

- maybe it's better this way
image
image

iwiederhold:
Why not limit the number?

M-Behr:
> > * [ ]   The "check" icon for the input field overlaps the arrows. Perhaps you can move it a little to the left.
> 
> This seems to be the first regular number input field we have. So I have different proposals for the check icon.
> 
> * move it simply a bit to the left could looks nasty with large numbers
>   image
>   image
> * maybe this way
>   image
>   image
> * or we use build a custom number input like for the quantity of cart products. But this seems to have no check. So there is the same problem like before.
>   image

I like the custom number input. We use it in order templates and quote requests too. There we display no "check" icon, which is fine for me. If a customer allows very large numbers, they have to adjust the field anyway.

M-Behr:
> Why not limit the number?

It's a cart configuration in the ICM. The customer can limit the amount of products in the cart. Our default is 100 but it should be possible to display larger numbers. With the custom input field we can display 7-digit numbers, which should be fine for now.

M-Behr (changes requested):
Nice work. It looks and behave better than before. I found some minor things:

- [x] The initial state of the quick order component is not displayed after adding product with "Add an item to Cart" section. The disabled cart button and the default qty of 1 are not set. It needs a reload of the page to fix it. 
2021-07-01 14_49_12-Your Shopping Cart _ Intershop PWA
- [x] The custom input field displays no error message and a "cross" icon in error case. Please make the behavior consistent to the other custom input fields.
quickorder_cart_error

SGrueber:
for the quantity input the <ish-product-quantity> component should be used, only this way it is the same behavior and look and feel all over the whole application. A combined input (variation and quantity) you can find on the product detail page

cbratsch:
> for the quantity input the ish-product-quantity component should be used, only this way it is the same behavior and look and feel all over the whole application. A combined input (variation and quantity) you can find on the product detail page

I changed the component and use now the product-quantity component in the form.
There is still an issue. The quantity field shows an error for the initial state and an invalid sku. Actually I have no good idea how to fix this issue. Maybe you can take a look on this or @dhhyi can give me a hint on this.

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -0,0 +1,122 @@
...
+import { GenerateLazyComponent } from 'ish-core/utils/module-loader/generate-lazy-component.decorator';
+
+@Component({
+  selector: 'ish-direct-order',
+  templateUrl: './direct-order.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+@GenerateLazyComponent()
+export class DirectOrderComponent implements OnInit, OnDestroy {
+  model = { sku: '', quantity: 1 };
```
SGrueber, line 20: quantity must not be part of the model

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.setContext();
+    this.directOrderForm.valueChanges
+      .pipe(takeUntil(this.destroy$))
+      .subscribe((control: { sku: string; quantity: number }) => {
+        this.context.set('sku', () => control.sku);
+        this.context.set('quantity', () => control.quantity);
+        this.model.quantity = control.quantity;
+        this.checkDisabledStates();
+      });
+    this.context
```
SGrueber, line 85: this is redundant

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  private setContext() {
+    this.context.set('sku', () => ' ');
+    this.context.set('minQuantity', () => 1);
+    this.context.set('maxQuantity', () => 100);
+    this.context.set('stepQuantity', () => 1);
+    this.context.set('hasQuantityError', () => false);
+  }
+
+  onSubmit() {
+    this.shoppingFacade.addProductToBasket(this.model.sku, this.model.quantity);
```
SGrueber, line 109: use addProductToBasket of the productContext facade

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -0,0 +1,122 @@
...
+      });
+  }
+
+  ngOnDestroy() {
+    this.destroy$.next();
+    this.destroy$.complete();
+  }
+
+  private setContext() {
+    this.context.set('sku', () => ' ');
```
SGrueber, line 101: this is not working at all because empty sku s are currently filtered in the product context facade

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -0,0 +1,122 @@
...
+    {
+      key: 'sku',
+      type: 'ish-text-input-field',
+      className: 'col-12 list-item search-container mb-1',
+      templateOptions: {
+        placeholder: 'shopping_cart.direct_order.item_placeholder',
+        autocomplete: true,
+      },
+      expressionProperties: {
+        'templateOptions.required': () => (!this.directOrderForm.untouched ? !!this.model.quantity : false),
```
SGrueber, line 36: read the quantity and the quantityError from the context facade, see product-add-to-basket component this.hasQuantityError$ = this.context.select('hasQuantityError');

SGrueber (changes requested):

M-Behr (approved):
It works and looks great - even better than in responsive starter store 👍 
Please still fix the little typo in the error message. Thanks a lot!

M-Behr (dismissed):

SGrueber (approved):

PR #752, opened 2021-06-29
feat: quickorder rework

cbratsch:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Issue Number: Closes #

## What Is the New Behavior?
[ ] Formly Rework
[ ] Fix Bug: Reset all Fields and add lines leads to broken design of new lines
[ ] check responsive behavior
[ ] Form Validation
[ ] min/ max Values
nice to have [ ] search suggest

[ ] Yes
[n] No

## Other Information

[AB#65470](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/65470)


M-Behr (changes requested):
The behavior is fine for the quick order. I would like to change some things:

- [x] Remove the tooltip
- [ ] Link: "Add another row" or "Add one more line"
- [ ] I would like to put the 2 "Add..." links closer together. We could use a | instead of space, e.g. Add another row | Add 5 more rows. We used this pattern on checkout for promotions "Details | Remove". Please discuss this with @iwiederhold.
2021-07-26 13_17_30-Intershop Progressive Web App _ Intershop PWA

cbratsch:
> The behavior is fine for the quick order. I would like to update some texts:
> 
> * [ ]  Tooltip: "Remove Row"
> * [ ]  Link: "Add another row" or "Add one more line"

The "tooltip is just a title. It's a bit hard to translate in this place and I think we don't really have this pattern somewhere else. Maybe we can just remove it?

Can you please provide also the german and french text for the link?

M-Behr:
> > The behavior is fine for the quick order. I would like to update some texts:
> > 
> > * [ ]   Tooltip: "Remove Row"
> > * [ ]   Link: "Add another row" or "Add one more line"
> 
> The "tooltip is just a title. It's a bit hard to translate in this place and I think we don't really have this pattern somewhere else. Maybe we can just remove it?
> 
> Can you please provide also the german and french text for the link?

You are absolutely right. For links, a tooltip makes no sense at all. We use this pattern only for icons. Please remove the tooltip for the link :) I will edit my review comment.
@marschmidt89 Please provide the German & French text for the link "Add another row" or "Add one more line". You can pick one of them in English. I want to make clear that only one line/row is added with this action (see screenshot above).

src/app/extensions/quickorder/shared/quickorder-add-products-form/quickorder-add-products-form.component.html
```diff
@@ -0,0 +1,21 @@
...
+    <div class="col-md-6 list-header-item no-seperator">{{ 'quickorder.page.product.id' | translate }}</div>
+    <div class="col-md-2 list-header-item no-seperator">{{ 'quickorder.page.product.quantity' | translate }}</div>
+  </div>
+
+  <formly-form [model]="model" [fields]="fields" [options]="options" [form]="quickOrderForm" class="pt-1"></formly-form>
+
+  <button
+    type="submit"
+    class="btn btn-primary"
+    [disabled]="quickOrderForm.pristine || quickOrderForm.invalid"
```
cbratsch, line 12: @SGrueber
I removed the check for quickOrderForm.invalid, because the form validation is somehow broken from the async validator in combination of the repeating.
So this disable or enable the button in an incomprehensible way.

M-Behr:
Here are the localizations for the link text:

- Add one more line
- Eine weitere Zeile hinzufügen
- Ajouter une ligne supplémentaire

Thank you @marschmidt89 for the translation!

M-Behr (changes requested):
I think it looks and works very well. I found some little things:

- [x] When I click "Add another row" or "Add 5 more rows", the "Add to Cart" button is enabled. I would suggest that click on these actions has no effect on "Add to Cart" button.
- [x] Placeholder "Product ID" is needed for responsive view because the labels are not visible.
- [x] Horizontal form is too large for small devices. We could use "Trash" icon instead of "Remove" link or make the input field smaller. @iwiederhold what do you think? See screenshot below.
- [x] The "|" link separator should not be part of the link. It should have normal text style. See example in promotions.

Example - Promotions:
2021-07-28 11_14_15-Your Shopping Cart _ Intershop PWA

Horizontal form in phone view:
2021-07-28 11_26_47-Intershop Progressive Web App _ Intershop PWA

cbratsch:
> * [ ]  When I click "Add another row" or "Add 5 more rows", the "Add to Cart" button is enabled. I would suggest that click on these actions has no effect on "Add to Cart" button.

We can create a known issue on this. Actually the form is only checked to be pristine. When a line is added or removed the form is changed and not pristine anymore. We need to check this with validity. That's not possible at the moment


> * [ ]  The "|" link separator should not be part of the link. It should has normal text style. See example in promotions.

The example in promotions isn't quite good. The separator sign should not be part of the html. We want to adapt this. But I put the css-class to an extra empty span, so it's not part of the link anymore.

M-Behr:
> > * [ ]   When I click "Add another row" or "Add 5 more rows", the "Add to Cart" button is enabled. I would suggest that click on these actions has no effect on "Add to Cart" button.
> 
> We can create a known issue on this. Actually the form is only checked to be pristine. When a line is added or removed the form is changed and not pristine anymore. We need to check this with validity. That's not possible at the moment

It works fine when I remove a row. However, it is fine if this task is solved later.

> > * [ ]   The "|" link separator should not be part of the link. It should has normal text style. See example in promotions.
> 
> The example in promotions isn't quite good. The separator sign should not be part of the html. We want to adapt this. But I put the css-class to an extra empty span, so it's not part of the link anymore.

Oh, okay. Yes, please also adapt the code for the promotion example later. I didn't know that the implementation wasn't that good. In any case, it should then look like normal text (no interactions for "|").

M-Behr (approved):
Much better than our Responsive Starter Store, especially the behavior in mobile and the nice quantity fields! 👍

SGrueber (approved):

SGrueber (approved):

PR #766, opened 2021-07-01
explore: what if the ICM API could fetch multiple products at once?

### LINKED ISSUE #764: provide shopping facade method that can fetch the product information for multiple products

It would be nice for projects if our `shopping.facade.ts` would already provide a method that could fetch the product information for multiple product SKUs in one step, e.g. `products$(skus: string[])`.

Currently one has to trigger `product$(sku: string | Observable<string>, level: ProductCompletenessLevel | true)` for each product of interest individually within a loop over the product SKUs array.

A current implementation would internally require to trigger individual products REST calls as long as ICM does not provide a REST API that would return the product information for multiple products requested by SKU. But this should not prevent working on this issue and refactoring the facade method once ICM provides an optimized REST API.

dhhyi:
## PR Type

[x] Other: Showcase

Feature request #764

## What Is the New Behavior?

Solution if the API could actually fetch multiple products at once.

[ ] Yes
[x] No

[AB#118537](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/118537)

### DIFF
--- a/src/app/core/services/products/products.service.ts
+++ b/src/app/core/services/products/products.service.ts
@@ -1,7 +1,7 @@
 import { HttpParams } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Observable, of, throwError } from 'rxjs';
-import { defaultIfEmpty, map, switchMap } from 'rxjs/operators';
+import { catchError, defaultIfEmpty, map, switchMap } from 'rxjs/operators';
 
 import { AttributeGroupTypes } from 'ish-core/models/attribute-group/attribute-group.types';
 import { CategoryHelper } from 'ish-core/models/category/category.model';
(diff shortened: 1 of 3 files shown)


dhhyi:
@shauke I think this would actually be a nice feature to have for the ICM API. It shouldn't be any API break either 😉

shauke:
Waiting for ICM internal request to provide a multiple SKUs REST API call:
https://dev.azure.com/intershop-com/Products/_workitems/edit/64436
https://dev.azure.com/intershop-com/Products/_workitems/edit/54822

shauke:
Closed since ICM does not seem to get the necessary REST request that could fetch multiple product details with one request.

PR #779, opened 2021-07-09
SSL certificate error message

### LINKED ISSUE #756: Provide better message when ICM uses self-signed SSL cert
**Actual Behavior**

When ICM backend is not protected with a valid SSL certificate the SSR rendering process logs a not very helpful message into std-out.
`ERROR Http failure response for https://somehost.intershop.de/INTERSHOP/rest/WFS/XYZ-Site/-;loc=en_US;cur=USD/configurations: 0 Unknown Error`

**Expected Behavior**

Please log some information into SSR std-out that SSL trust cannot be established.

1. Select some ICM_BASE_URL value that fits the SSL certificate requirements
2. Deploy and start containers with `docker-compose up`
3. Open browser to see homepage
4. See error in error log of docker-compose process

[AB#64626](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/64626)

Eisie96:
## PR Type

[x] Bugfix

When ICM backend is not protected with a valid SSL certificate the SSR rendering process logs a not very helpful message into std-out.

Issue Number: Closes #756

## What Is the New Behavior?

Log some information into SSR std-out that SSL trust cannot be established.

[x] No

## Other Information

[AB#64704](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/64704)

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -14,8 +14,6 @@ const PORT = process.env.PORT || 4200;
 
 const DIST_FOLDER = join(process.cwd(), 'dist');
 
-const ICM_BASE_URL = process.env.ICM_BASE_URL || environment.icmBaseURL;
-
 // uncomment this block to prevent ssr issues with third-party libraries regarding window, document, HTMLElement and navigator
 // tslint:disable-next-line: no-commented-out-code
 /*
@@ -35,6 +33,8 @@ global['navigator'] = win.navigator;
 export function app() {
   const logging = /on|1|true|yes/.test(process.env.LOGGING?.toLowerCase());
 
+  const ICM_BASE_URL = process.env.ICM_BASE_URL || environment.icmBaseURL;
+
   if (!ICM_BASE_URL) {
     console.error('ICM_BASE_URL not set');
     process.exit(1);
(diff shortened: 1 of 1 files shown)


server.ts
```diff
@@ -33,7 +35,6 @@ global['navigator'] = win.navigator;
 export function app() {
   const logging = /on|1|true|yes/.test(process.env.LOGGING?.toLowerCase());
 
-  const ICM_BASE_URL = process.env.ICM_BASE_URL || environment.icmBaseURL;
```
jometzner, line 36: Moving this to a larger scope means that it lacks a "falsey" check at the right time. You are accessing this value in function `run()` but the falsey check is performed in `app` function. To late...

jometzner (changes requested):

server.ts
```diff
@@ -263,8 +264,53 @@ export function app() {
...
+      'CERT_REVOKED',
+      'INVALID_CA',
+      'INVALID_PURPOSE',
+      'CERT_UNTRUSTED',
+      'CERT_REJECTED',
+      'HOSTNAME_MISMATCH',
+    ];
+
+    req.on('error', e => {
+      if (certErrorCodes.indexOf(e.code) > -1) {
```
dhhyi, line 301: [Array.includes](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

server.ts
```diff
@@ -263,8 +264,53 @@ export function app() {
 }
 
 function run() {
+  const https = require('https');
```
dhhyi, line 267: Shouldn't this check only be done if the ICM_BASE_URL actually starts with https? This check will always fail if I use a local URL without TLS

jometzner (approved):

jometzner (dismissed):

jometzner (approved):

PR #782, opened 2021-07-12
DeployDemoServer for non intershop repos

svsaller:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ x] CI-related changes
[ ] Other: 

Issue Number: Closes #

[ ] Yes
[ ] No

### DIFF
--- a/.github/workflows/demo-server.yml
+++ b/.github/workflows/demo-server.yml
@@ -22,7 +22,7 @@ jobs:
 
   BuildNDeployDemoServer:
     needs: [CancelPrevious]
-    if: github.event.pull_request.head.repo.owner.login == 'intershop'
+    # if: github.event.pull_request.head.repo.owner.login == 'intershop'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout


dhhyi:
I think, this should not be working, as the needed Azure AppService credentials are not shared to forked repositories.

svsaller:
Yes @dhhyi, I had expected that as well. Could this be made available (at least technically) to specific contributors to make the demo server deployment work for them?

dhhyi:
> Could this be made available (at least technically) to specific contributors to make the demo server deployment work for them?

That would mean exposing the docker registry credentials and also the service principal credentials. I think I can safely respond with "NO!" 😄

You can ask for an invite as contributor to the project and commit directly to it instead of your own fork.

Any other thoughts @jometzner @shauke @rkarl-ish ?

shauke:
I found the condition as well but I also read about the GitHub secrets not being shared to the fork pull requests. So as @dhhyi already mentioned this would not solve the problem. So we need to think about inviting the developers as direct contributors or we need to copy branches and pull requests in our repository that need a review server.

jometzner:
It is quite simple:
- You want demo server capabilities in your fork? Please, built a workflow that builds and deploys your artifacts in your environment. 
- You want to use our demo environment? Kindly ask us to become a contributor to this project.

PR #794, opened 2021-07-23
Fix/add missing quickorder link in mobile view

cbratsch:
## PR Type

[x] Bugfix

The quick order link is missing in phone (xs) view. This means no access to the feature for phone devices.

## What Is the New Behavior?
The quick order link is visible in phone view. It is placed in the hamburger menu (see IAD).
image

[ ] Yes
[x] No

## Other Information

[AB#65435](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/65435)


src/app/shell/header/header-default/header-default.component.html
```diff
@@ -73,7 +73,9 @@
       </div>
     </div>
     <div class="col-md-5 text-right">
-      <ish-lazy-header-quickorder></ish-lazy-header-quickorder>
+      <div class="d-none d-md-block" *ishFeature="'quickorder'">
```
shauke, line 76: It is nice to have a consistent pattern in regards to applying directives and CSS classes etc. And since the conditional directive is mostly more important it is better to read if it is in first place (see the other changes).

`<div *ishFeature="'quickorder'" class="d-none d-md-block">`

It reads much cleaner "render only if quickorder feature is enabled" and only in this case the CSS classes are relevant.

src/app/extensions/quickorder/shared/header-quickorder/header-quickorder.component.html
```diff
@@ -1,4 +1,6 @@
-<div class="d-none d-md-block quick-order-links">
-  <fa-icon class="mr-2 text-primary align-middle" [icon]="['fas', 'fast-forward']"></fa-icon>
-  <a rel="quick-order" routerLink="/quick-order">{{ 'quickorder.page.link' | translate }}</a>
-</div>
+<fa-icon
+  class="mr-2 align-middle"
```
shauke, line 2: Maybe we could remove the usage of Bootstrap Utility Classes again? This is a paradigm change that would make customizations different.

shauke (changes requested):
- [x] There styling in mobile view does not match the other links
image

- [x] The current rendering logic does not work for the Sticky Header
image

- [x] It would be nice to still handle the feature toggle check only in the lazy component and also handle the `div` rendering in the `header-quickorder.component.html` itself so one can still use only `<ish-lazy-header-quickorder></ish-lazy-header-quickorder>` where ever needed. The additional input is fine. Internally `ng-template` (see `product-image.component.html` could be used).

M-Behr (changes requested):
- [ ] Space between icon and quick order link in mobile hamburger menu.
- [ ] Remove quick order link from Sticky header.

M-Behr (commented):
I noticed that the hover effect of the quick order link is inconsistent with the other links in hamburger menu. Intuitively, I would say yours is right and all the others are incorrect. 

I discussed this with @iwiederhold and your styling is correct. Maybe the other link hover styles are removed with this CSS purge thing? Can you match the other link hover styles to yours or do we make an extra ticket for that?

shauke (changes requested):

M-Behr:
> I noticed that the hover effect of the quick order link is inconsistent with the other links in hamburger menu. Intuitively, I would say yours is right and all the others are incorrect.
> 
> I discussed this with @iwiederhold and your styling is correct. Maybe the other link hover styles were removed with this CSS purge thing? Can you match the other link hover styles to yours or do we make an extra ticket for that?

I tracked a minor bug for this VD issue: #805

M-Behr (changes requested):
I found an additional blank character in "Quick Order". This results in the link not being in line with the other links:
QuickOrderLink

shauke:
> 
> I found an additional blank character in "Quick Order". This results in the link not being in line with the other links:

This is fixed with the latest changes.

src/app/shell/header/user-information-mobile/user-information-mobile.component.html
```diff
@@ -4,6 +4,9 @@
   <div *ishFeature="'compare'" class="user-info-box-item">
     <ish-product-compare-status view="full"></ish-product-compare-status>
   </div>
+  <div *ishFeature="'quickorder'">
```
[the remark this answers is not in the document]
shauke: Changed to `<div *ishFeature="'quickorder'" class="user-info-box-item">` to be consistent with the other items. It is not quite the same as in `src/app/shell/header/header-default/header-default.component.html` as mentioned earlier.

M-Behr:
> > I found an additional blank character in "Quick Order". This results in the link not being in line with the other links:
> 
> This is fixed with the latest changes.

Hmm, but I can still see it. When I remove the blank character with the browser tool, it looks fine. I also tested it with an incognito window. Does it no longer occur in your local environment? Maybe it has something to do with the review server.

shauke (dismissed):

M-Behr (dismissed):
Perfect 👍

PR #818, opened 2021-08-03
feat: readiness and liveness probe for pm2

Eisie96:
## PR Type

[x] Feature

K8s uses a liveness and readiness probes to check if a pod is available and ready to accept requests. Due to the implemented pm2 process manager a new liveness and readiness probe is needed.

## What Is the New Behavior?

New scripts for liveness and readiness probes are added to project, which can be used in k8s deployments.
The readiness probe checks if at least one process is available for each theme.
The liveness probe checks for each theme, that at least one process with status 'online' or 'launching' exists. 

The healthcheck should not be used for liveness and readiness probes and should not check, if the ICM is available.

[ x ] No

## Other Information

[AB#65808](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/65808)

### DIFF
--- a/dist/livenessprobe.js
+++ b/dist/livenessprobe.js
@@ -1,0 +1,34 @@
+const pm2 = require('pm2');
+let ports = require('./ecosystem-ports').ports;
+
+if (process.env.ACTIVE_THEMES) {
+  const active = process.env.ACTIVE_THEMES.split(',');
+  ports = Object.entries(ports)
+    .filter(([theme]) => active.includes(theme))
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {});
+}
+
+pm2.connect(err1 => {
+  if (!err1) {
+    pm2.list((err2, list) => {
+      if (!err2) {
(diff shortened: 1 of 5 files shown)


dist/livenessprobe.js
```diff
@@ -0,0 +1,34 @@
...
+        Object.entries(ports).forEach(([theme]) => {
+          if (
+            list
+              .filter(el => el.name === theme)
+              .find(el => el.pm2_env.status === 'online' || el.pm2_env.status === 'launching')
+          ) {
+            process.exit(0);
+          }
+        });
+        process.exit(1);
```
dhhyi, line 24: This will exit with 0 if the first process is found and not check other themes.

How about:
```
const allUp = Object.entries(ports).map(([theme]) => list...filter...find).reduce((acc, val) => acc && !!val, true);
process.exit(allUp ? 0 : 1);
```

dhhyi (changes requested):

jometzner (approved):

dhhyi (approved):

PR #841, opened 2021-08-19
feat: add virtual theme "all" for always overriding files

dhhyi:
## PR Type

[x] Feature
[x] Build-related changes

Currently it's not possible to _always_ override a file for every theme. All themes have to be added to the filename.

## What Is the New Behavior?

By using the virtual theme "all", a file is always switched in if no theme-specific override is defined.

[ ] Yes
[x] No

## Other Information

[AB#66222](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/66222)


dhhyi:
@shauke We are using it for a couple of weeks in a project now and no issues came up. 😊👍

shauke (approved):

PR #844, opened 2021-08-22
fix: prevent call for empty product on quickorder and direct order

dhhyi:
## PR Type

[x] Bugfix

Quickorder and direct order do calls to the product API with an empty product sku.

## What Is the New Behavior?

Fixed

[ ] Yes
[x] No

## Other Information

[AB#66292](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/66292)


cbratsch:
There seems to be a time issue on the quickorder page when the page is fresh loaded.
Steps to Repeat:
1. go to the quickorder page and reload the page there
2. fill in a sku and click on the '+'- button of the associated quantity field
-> the quantity stays at 1
-> Error message 'Please enter a whole number' shows up

dhhyi:
> 2. fill in a sku and click on the '+'- button of the associated quantity field

That works for me with the deployed demo server 🤔. Did you try it locally? Anything that I'm missing?

cbratsch:
I tested locally. Now I rebased, resolved the conflict and add another fix. When the build is finished, I'll check on demo server again.

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -44,7 +41,13 @@ export class DirectOrderComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   ngAfterViewInit() {
-    this.context.connect('sku', this.directOrderForm.get('sku').valueChanges);
+    this.context.connect(
+      'sku',
+      this.directOrderForm.get('sku').valueChanges.pipe(
+        tap(() => (this.loading = true)),
```
dhhyi, line 47: I think this can be refactored to use `context.select('loading')` 😅

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
```diff
@@ -44,7 +41,13 @@ export class DirectOrderComponent implements OnInit, OnDestroy, AfterViewInit {
...
+        debounceTime(500)
```
dhhyi, line 48: double requests? Did you check out `distinctUntilChanged`?

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
dhhyi: 🥰

src/app/extensions/quickorder/shared/direct-order/direct-order.component.ts
cbratsch: One Request came from the context. The other from the check of the asnyc validator, which asked the shopping facade. Same issue is in the quickorder page. I actually try to fix this there too.

M-Behr (changes requested):
The behavior is fine. I found a little issue with master variation products.
**Steps to Repeat:**
1. Enter SKU 201807231 => Quantity field is removed which is ok because you can't add this product.
2. Click "Add to Cart" => A product variation 201807231-04 is added to cart.

The product variation should not be added. The quantity is 0 because quantity field is not visible. So nothing should be added for this SKU.
Alternativ: We display an error message in mini cart (same behavior like for invalid SKUs). But in that case, the quantity field should not disappear.

cbratsch:
> The behavior is fine. I found a little issue with master variation products.
> 
> Steps to Repeat:
> 
> 1. Enter SKU 201807231 => Quantity field is removed which is ok because you can't add this product.
> 2. Click "Add to Cart" => A product variation 201807231-04 is added to cart.
> 
> The product variation should not be added. The quantity is 0 because quantity field is not visible. So nothing should be added for this SKU. After click on "Add to Cart", the SKU field is empty and quantity field should come up again.
> Alternativ: We display an error message in mini cart (same behavior like for invalid SKUs). But in that case, the quantity field should not disappear.

I wrote a feature request (#847). Requirements are needed for this. This shouldn't be handled in this fix branch.

M-Behr (approved):

MaxKless (approved):

PR #853, opened 2021-08-27
refactor: build and bundle docker scripts with webpack

dhhyi:
## PR Type

[x] Build-related changes

Current docker build installs the currently available versions of `express`, `express-http-proxy` `pm2` and `prom-client` without pinning versions for repeatable builds.

## What Is the New Behavior?

- Scripts are bundled with webpack to pin down versions.
- `pm2` and `express` are installed pinned down into the container.

[X] Yes
[ ] No

## Other Information

[AB#69315](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69315)


jometzner:
I see warnings in the console. Can you tell me how to get rid of those? Webpack configuration tweak?
```console
WARNING in ./node_modules/express/lib/view.js 81:13-25
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/express/lib/application.js 22:11-28
 @ ./node_modules/express/lib/express.js 18:12-36
 @ ./node_modules/express/index.js 11:0-41
 @ ./src/ssr/server-scripts/server.js 5:16-34
```

dhhyi:
> I see warnings in the console. Can you tell me how to get rid of those? Webpack configuration tweak?

Well, the problem is just that webpack can't resolve a dynamical `require` in express.js. That does not seem to be really fixable. And I also would not like to omit this warning.

One solution would be to also externalize `express` and install it in the container like pm2, but then, the version would not be pinned down anymore.
Another solution would be to also write a `package.json` into the dist folder and do a `npm install` when building. I can try that.

dhhyi:
@jometzner Idea No. 2 seems to be working 😊

jometzner:
Solution (1) I also tried but skipped as pinning was the whole point of the exercise. Solution (2) did not cross my mind. I'll have a look at that today.

jometzner:
@dhhyi In my opinion we need a migration note for this change. Have a look at my additional commits.

dhhyi:
> @dhhyi In my opinion we need a migration note for this change. Have a look at my additional commits.

Yes, good idea 👍

jometzner (approved):

jometzner (approved):

PR #859, opened 2021-09-01
no broken page on missing requisition data

DiverDori:
## PR Type

[x] Bugfix

There might be some data of a requisition not available, especially if the user got deleted. This should not break the page.

## What Is the New Behavior?

page stays intact

[x] No

## Other Information

[AB#69421](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69421)


shauke (approved):
Good job. 👍

PR #861, opened 2021-09-06
fix: enable cxml punchout with unclean user state

MaxKless:
## PR Type

[x] Bugfix

If you are logged into the PWA, there will some kind of persistent user state saved to your browser, depending on the identity provider. Especially with SSO this can lead to all kinds of problems if you are logged into the PWA and then try to do cxml punchout in the same browser.
This PR will attempt to fix this behaviour so that cxml punchout works even if the user forgot to log out of the regular PWA or has some other uncleaned state in their browser.

## What Is the New Behavior?

Punchout will work regardless of persistent user state.

[x] No

## Other Information

[AB#69525](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69525)

### DIFF
--- a/src/app/core/identity-provider/identity-provider.interface.ts
+++ b/src/app/core/identity-provider/identity-provider.interface.ts
@@ -39,7 +39,7 @@ export interface IdentityProvider<ConfigType = never> {
   /**
    * Route guard for logout
    */
-  triggerLogout(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): TriggerReturnType;
+  triggerLogout(route?: ActivatedRouteSnapshot, state?: RouterStateSnapshot): TriggerReturnType;
 
   /**
    * Interceptor for all API requests directed to the ICM
--- a/src/app/core/identity-provider/identity-provider.interface.ts
+++ b/src/app/core/identity-provider/identity-provider.interface.ts
@@ -37,9 +37,9 @@ export interface IdentityProvider<ConfigType = never> {
   triggerInvite?(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): TriggerReturnType;
 
   /**
-   * Route guard for logout
+   * Logout method (can be used by route guard or otherwise)
    */
-  triggerLogout(route?: ActivatedRouteSnapshot, state?: RouterStateSnapshot): TriggerReturnType;
+  triggerLogout(): TriggerReturnType;
 
   /**
    * Interceptor for all API requests directed to the ICM
--- a/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
+++ b/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
@@ -7,6 +7,7 @@ import { catchError, concatMap, delay, first, mapTo, switchMap, take, tap } from
 import { AccountFacade } from 'ish-core/facades/account.facade';
 import { AppFacade } from 'ish-core/facades/app.facade';
 import { CheckoutFacade } from 'ish-core/facades/checkout.facade';
+import { IdentityProviderFactory } from 'ish-core/identity-provider/identity-provider.factory';
 import { ApiTokenService } from 'ish-core/utils/api-token/api-token.service';
 import { CookiesService } from 'ish-core/utils/cookies/cookies.service';
 import { whenTruthy } from 'ish-core/utils/operators';
@@ -23,6 +24,7 @@ export class PunchoutPageGuard implements CanActivate {
     private apiTokenService: ApiTokenService,
     private cookiesService: CookiesService,
     private punchoutService: PunchoutService,
+    private identityProvider: IdentityProviderFactory,
     @Inject(PLATFORM_ID) private platformId: string
   ) {}
 
@@ -48,6 +50,7 @@ export class PunchoutPageGuard implements CanActivate {
 
     // initiate the punchout user login with the access-token (cXML) or the given credentials (OCI)
     if (route.queryParamMap.has('access-token')) {
+      this.identityProvider.getInstance().triggerLogout();
       this.accountFacade.loginUserWithToken(route.queryParamMap.get('access-token'));
     } else {
       this.accountFacade.loginUser({
--- a/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
+++ b/src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
@@ -48,9 +48,9 @@ export class PunchoutPageGuard implements CanActivate {
       return this.router.parseUrl('/loading');
     }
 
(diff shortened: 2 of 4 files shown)


src/app/core/identity-provider/identity-provider.interface.ts
```diff
@@ -39,7 +39,7 @@ export interface IdentityProvider<ConfigType = never> {
   /**
    * Route guard for logout
    */
-  triggerLogout(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): TriggerReturnType;
+  triggerLogout(route?: ActivatedRouteSnapshot, state?: RouterStateSnapshot): TriggerReturnType;
```
dhhyi, line 42: this method is meant to be (as documented) a route guard and not to be called directly. Is it possible, that instead of calling it, you just navigate to `/logout`?

dhhyi (changes requested):

src/app/core/identity-provider/identity-provider.interface.ts
shauke: Navigating to `/logout` would not be helpful since we want to navigate to a different route e.g. `/home`. Only the logged in state of the user should be cleaned up so the intended login with the punchout user works as intended on the way to `/home` or any other intended route not being `/logout`.

src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
```diff
@@ -48,6 +50,7 @@ export class PunchoutPageGuard implements CanActivate {
 
     // initiate the punchout user login with the access-token (cXML) or the given credentials (OCI)
     if (route.queryParamMap.has('access-token')) {
+      this.identityProvider.getInstance().triggerLogout();
       this.accountFacade.loginUserWithToken(route.queryParamMap.get('access-token'));
     } else {
       this.accountFacade.loginUser({
```
shauke, line 56: Is the logout to clean up previous user session information only needed for cXML punchout login (with access token) or wouldn't this be needed for the OCI punchout with username and password as well?

src/app/core/identity-provider/identity-provider.interface.ts
dhhyi: So you want to navigate to `/logout?returnUrl=/home`? 😅

src/app/extensions/punchout/pages/punchout/punchout-page.guard.ts
MaxKless: You're right, I modified the behavior to include OCI

src/app/core/identity-provider/identity-provider.interface.ts
shauke: We do not just want to navigate to `/home`, we also want to login the punchout user. This is the more important part. So we might need something like `/logout?returnUrl=/punchout?USERNAME=<USERNAME>&PASSWORD=<PASSWORD>&HOOK_URL=<HOOK_URL>`. It does not really look good to me.
So the problem is more that we use an internal function of the route guard. Maybe we could clean this up?

dhhyi: If I look at the code where the `triggerLogout()` is called, isn't it something else you want to call? Like `this.apiTokenService.removeApiToken()` perhaps?

Please correct me if I'm wrong: Punchout uses the default "old" ICM login mechanism that works with the API token. So we actually know, that all we want to do is resetting the token, right?

I still think making the arguments optional and calling this guard method directly is a wrong approach.

dhhyi: I just had another look at the description of the ticket. So you really need to bypass SSO state. Why not just always have a callback URL, that also has `;identityProvider=ICM` in it? That way the SSO state will be ignored for punchout, and only the ICM login method is valid. Furthermore, the SSO state will be properly reused, when using the page without punchout. @MaxKless @shauke

dhhyi: Alternatively, please go the current way the full way:

- remove the arguments from the `triggerLogout` method completely instead of making them optional
- add a `BREAKING CHANGE` comment

As far as I can see, the ICM and Auth0 IdentityProviders don't use them, and our AAD IdentityProvider in the project also doesn't need them.

MaxKless: Hi @dhhyi,
thanks for the review and lively discussion. You raise some good points here and it's true my implementation was abusing the `triggerLogout()` method. 
Sadly, the redirecting ideas didn't appeal to us very much either for various reasons and we decided to go the current way but cleaner.
- The identity provider interface is extended with a `logout` method that is meant to give a programmatic way of logging out a user without using the `/logout` route
- The `triggerLogout` methods have been refactored to use the `logout` method internally as well as do any extra routing logic
- The `punchout` guard uses the dedicated `logout` method
- The `triggerLogout` signature now has only required parameters again - it will only be used by guards.
- We will document this breaking change accordingly (every new idp needs to provide a `logout` method).

dhhyi: @MaxKless I think that is making things worse, sorry for my honest opinion 😅 

Now we have three methods of logging out a user, the latter which even has to be implemented by other implementors. Please reconsider and just use the direct call to `triggerLogout()` and remove all arguments. As I stated previously, that's an acceptable not-so-much conflicting API break.

MaxKless: Don't worry, good to have your opinion.

Yes, the api for identity providers becomes larger. But what happens if someone does end up needing the route information? Is it an acceptable drawback that they have to gather up the information themselves via the router?

dhhyi: They can get it in the router state, like it's actually done at the moment 😉 
https://github.com/intershop/intershop-pwa/blob/1fa76d1e25191cf4ccdbe4141c45ef2e63656933/src/app/core/identity-provider/icm.identity-provider.ts#L46-L50

MaxKless: Yeah that's what I meant... I guess it might not be too bad then. Just seems weird to have the signatures of the trigger methods different.

dhhyi: That's just a matter of JSDoc. The words `Route guard` are no longer valid for this method 😅

dhhyi: May I ask what the "various reasons" were for not going for the redirecting solution? I think it would actually the best solution, to have a dedicated Identity Provider for the whole punchout feature in the first place.

dhhyi (approved):

dhhyi:
@MaxKless don't forget the `BREAKING CHANGE` when merging 😉

PR #869, opened 2021-09-13
feat: HTTP request creator

USSDlover:
Update ngrx-creator

Add HTTP request creator to be able to create three action in one line and decrease line of codes significantly.

## PR Type

[❌] Bugfix
[✔] Feature
[❌] Code style update (formatting, local variables)
[❌] Refactoring (no functional changes, no API changes)
[❌] Build-related changes
[❌] CI-related changes
[❌] Documentation content changes
[❌] Application / infrastructure changes
[❌] Other: 

Issue Number: Closes #

## What Is the New Behavior?
Create HTTP actions for three scenarios in one line

[❌] Yes
[✔] No

### DIFF
--- a/src/app/core/utils/ngrx-creators.ts
+++ b/src/app/core/utils/ngrx-creators.ts
@@ -1,4 +1,5 @@
-import { on } from '@ngrx/store';
+import { ActionCreator, createAction, on } from '@ngrx/store';
+import { TypedAction } from '@ngrx/store/src/models';
 
 import { HttpError } from 'ish-core/models/http-error/http-error.model';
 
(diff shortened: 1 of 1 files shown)


USSDlover:
Let me just come back with another PR with unit test and solution for this TypedAction which imported from `@ngrx/store/src/models`
Also, I can see as the name of the payloads is not `payload` to prevent the shadow label lint rule, it making errors when used in actions.

MaxKless:
Okay sure, thanks for working to contribute @USSDlover. Maybe you can refactor a part of the code that this PR would help to simplify so we understand exactly what you're going for with this action.

PR #871, opened 2021-09-14
fix: noSSR PWA Docker container does not support deep links

shauke:
## PR Type

[x] Bugfix

After the latest changes to the noSSR PWA Docker container the container only worked for the '/' route but not for any other route anymore, e.g. '/home'. Such routes return a 404 error from nginx.

image

## What Is the New Behavior?

Reverted to the former location matching in nginx to restore previously working noSSR container routing.

[x] No

## Other Information

[AB#69756](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69756)

### DIFF
--- a/templates/nginx.conf
+++ b/templates/nginx.conf
@@ -34,13 +34,9 @@ http {
         listen  [::]:4200;
         server_name  localhost;
 
-        location ~* /.*index.html {
-            root   /usr/share/nginx/html;
-            try_files $uri $uri/ /index.html;
-        }
-
         location / {
             root   /usr/share/nginx/html;
+            try_files $uri $uri/ /index.html;
 
             if ($request_method = OPTIONS) {
                 return 204;


dhhyi (dismissed):
That is interesting, the `try_files` comes from here: https://angular.io/guide/deployment. I remain sceptic...

dhhyi:
two more thoughts: remove support for no SSR and service worker. It will clean up the code base quite a bit. The SSR image on PWA 1.1 runs pretty smooth. There is no need to support any shenanigans @shauke @jometzner @rkarl-ish

dhhyi:
> That is interesting, the `try_files` comes from here: https://angular.io/guide/deployment. I remain sceptic...

@shauke 
Ah, you moved it, now I see.

shauke:
Yes, I basically reverted the location matching changes introduced with your last commit.
The relevant part is `location ~* /.*index.html {` back to `location / {` that worked as expected.

Regarding the `try_files $uri $uri/ /index.html;` part I hope it works correct together with the additions for the header at the used place in code.

shauke:
And yes you are right we should think about the usefulness of the noSSR container and the Service Worker as well. Good hint.
But this PR is just to make the noSSR container functional again.

PR #874, opened 2021-09-16
deps: update to Angular 12.2.6 and dependencies updates

shauke:
## PR Type

[x] Other: dependencies updates

## To Dos

- [x] project no longer compiles after `@angular/cli` update - (pluralization issue?)
  - `raw-loader` was missing. Needed for `data-testing-*` removal. 
- [x] can `postinstall` be updated to 0.7.3 with code adaptions?
  - complete paths have to be supplied now 

[x] No

## Other Information

[AB#69872](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69872)

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install commitlint --edit "$1"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged --concurrent false
--- a/.prettierignore
+++ b/.prettierignore
@@ -21,6 +21,7 @@
 *.crt
 *.key
 .*ignore
+.husky
 .gitattributes
 .editorconfig
 .createMockdata.table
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -5,6 +5,7 @@ COPY schematics /workspace/schematics/
 COPY package.json package-lock.json /workspace/
 RUN npm i --ignore-scripts
 RUN npm run ngcc
+RUN node node_modules/esbuild/install.js
```
dhhyi, line 8: @shauke @MaxKless this should actually be part of the `npm postinstall` as simplified in #865

shauke: Does this mean the other PR needs adaptions once this PR is merged together with the "chore: fix docker build" commit or what is your suggestion?

dhhyi: Yes, like that

dhhyi (changes requested):
The two `chore: fix` commits 785e8be316cead97b6a0cb33b0fbd64f4f3591a8 and 6cfa7bc5416e43ac701da2b7811c09a895317012 should be amended to the upgrade commit 273f6d146e396c03ba7510111a916d11d18a0cd1, so that each commit builds consistently.

Can you also shorten the headline of 273f6d146e396c03ba7510111a916d11d18a0cd1 and move the packages into the body of the commit message? 😃

dhhyi (approved):

PR #876, opened 2021-09-18
fix: use translate get instead of instant for product-image component

dhhyi:
## PR Type

[x] Bugfix

Occasionally the console output contains the error `missing translation in en_US: product.image.text.alttext`. This happens because the `product-image` component accesses this translation with `TranslateService.instant` when translations are not yet loaded correctly.

## What Is the New Behavior?

Refactored to use `TranslateService.get`

[ ] Yes
[x] No

## Other Information

[AB#69899](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69899)


MaxKless (approved):
works as intended and fixes the issue :)

PR #881, opened 2021-09-19
perf: cache compiled translations in PWA process

dhhyi:
## PR Type

Translate Compiles are re-done for every value when the translations are initialized. Especially in the SSR process caching can improve performance.

## What Is the New Behavior?

Once compiled, translation templates are cached and reused.

[ ] Yes
[x] No

## Other Information

[AB#69905](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69905)

### DIFF
--- a/src/app/core/utils/translate/pwa-translate-compiler.ts
+++ b/src/app/core/utils/translate/pwa-translate-compiler.ts
@@ -3,6 +3,8 @@ import { TranslateCompiler, TranslateService } from '@ngx-translate/core';
 
 import { Translations } from './translations.type';
 
+const cache: Record<string, Function> = {};
+
 @Injectable()
 export class PWATranslateCompiler implements TranslateCompiler {
   constructor(private injector: Injector) {}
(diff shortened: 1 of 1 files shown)


MaxKless:
Hi, @dhhyi thanks for the great PR. I found an issue, could you maybe help me understand what is happening. Sometimes, the cache resets itself and I can't really figure out why. 
Steps to repeat:
- Add some kind of logging to the compile function (so you can see what's in the cache object)
- Start a local dev environment (with nginx & ssr)
- load the home page -> see the cache fill up
- reload the page
-> Sometimes the cache is full (11 entries per default) and sometimes the cache starts empty and fills up again during the reload. I can't really find a pattern to this. 

Any idea?

dhhyi:
Hi @MaxKless,

> * Start a local dev environment (with nginx & ssr)
> ...
> Any idea?

Did you start the SSR with the docker image or via docker-compose? If yes, then by default two processes are spawned and load balanced, so if you see a second process starting fresh, that is expected.
I can't reproduce the described behavior with just one single `npm run start`.

MaxKless (approved):

Issue #849, opened 2021-08-25
Disable checkout button when cart is not valid on cart page


M-Behr:
**Actual Behavior**
"Checkout" button is enabled when cart is invalid (e.g. no cost center selected, invalid quantity). An error message appears, but button is enabled.

**Expected Behavior**
"Checkout" button is disabled until user do something in the cart (e.g. delete a product, update quantity, select cost center).
It should have the same behavior like in other checkout steps, e.g. payment page.

Note: Invalid promotion code have no effect on "Checkout" button.

**Steps to Reproduce the Bug**

1. Add product to cart
2. Enter "-200" in quantity or set cost center to "Please select"
3. Click on "Checkout"
4. Go to payment step
5. Select no payment method
6. Click "Continue Checkout"

The behavior is different. I would like to add the behavior on cart page. Payment step is fine.

[AB#69216](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69216)

PR #884, opened 2021-09-21
fix: restrict product-quantity input to numbers (browser dependend)

MoritzRS:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

When you enter negative numbers as product quantity, the form will display a warning and not set the value.

## What Is the New Behavior?
~~When you enter negative numbers as product quantity, they get inverted to their positive counterparts and the value can be set. This prevents negative numbers from being used in the first place as this doesn't make sense for quantity values.~~

Depending on the browser (Chromium based works, others not) the added input pattern `[0-9]*` prevents the adding of invalid characters (letters etc.) but not '-'. But this is handled via validation error.

[ ] Yes
[x] No

## Other Information

~~Possible solution to #849~~

### DIFF
--- a/src/app/shared/components/product/product-quantity/product-quantity.component.ts
+++ b/src/app/shared/components/product/product-quantity/product-quantity.component.ts
@@ -49,7 +49,7 @@ export class ProductQuantityComponent implements OnInit {
   }
 
   private setValue(value: number) {
-    this.context.set('quantity', () => value);
+    this.context.set('quantity', () => Math.abs(value));
   }
 
   private setNextValue(value: number) {
--- a/src/app/shared/components/product/product-quantity/product-quantity.component.ts
+++ b/src/app/shared/components/product/product-quantity/product-quantity.component.ts
@@ -49,7 +49,7 @@ export class ProductQuantityComponent implements OnInit {
   }
 
   private setValue(value: number) {
-    this.context.set('quantity', () => Math.abs(value));
+    this.context.set('quantity', () => value);
   }
 
   private setNextValue(value: number) {
@@ -69,5 +69,21 @@ export class ProductQuantityComponent implements OnInit {
 
   change(target: EventTarget) {
     this.setValue(Number.parseInt((target as HTMLDataElement).value, 10));
+  }
+
+  input(event: Event) {
+    const inputEvent = event as InputEvent;
+    const inputData = inputEvent.data;
+    const inputField = inputEvent.target as HTMLInputElement;
+    const allowed = /^[^0-9]*$/;
+    if (inputData !== null && allowed.test(inputData)) {
+      inputField.value = `${this.context.get('quantity')}`;
+    }
+  }
+
+  paste(event: Event) {
+    const clipboardData = (event as ClipboardEvent).clipboardData.getData('text');
+    const allowed = /^[0-9]*$/;
+    return allowed.test(clipboardData);
   }
 }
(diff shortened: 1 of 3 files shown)


dhhyi (approved):

MoritzRS:
I have an idea of a better way to fix this. Please don't merge yet as I want to test this other idea

dhhyi:
> I also just thought about the pattern based approach. Might try this tomorrow an create another PR

Just checking: https://caniuse.com/input-pattern It actually seems supported pretty well. So please, go ahead 👍

MoritzRS:
I added the pattern to the input element. But since pattern is only a guide for the browser, I also added additional functions to validate the input. This prevents the user from typing or pasting non numerical characters in the input field. However it does not check if the actual value of the input field is within the given boundaries. Because of this it does not qualify as a fix for #849

src/app/shared/components/product/product-quantity/product-quantity.component.ts
```diff
@@ -70,4 +70,20 @@ export class ProductQuantityComponent implements OnInit {
...
+    if (inputData !== null && allowed.test(inputData)) {
+      inputField.value = `${this.context.get('quantity')}`;
+    }
+  }
+
+  paste(event: Event) {
+    const clipboardData = (event as ClipboardEvent).clipboardData.getData('text');
+    const allowed = /^[0-9]*$/;
+    return allowed.test(clipboardData);
+  }
```
dhhyi, line 88: Instead of interrupting the `input` and `paste`, I think this can be done in the `product-context.facade.ts` by adding another validation, that the number must be positive:
https://github.com/intershop/intershop-pwa/blob/932473360d69774c2fa90649736e7638f1412469/src/app/core/facades/product-context.facade.ts#L191-L199

That way it will also become a fix for #849 again 😃

dhhyi (changes requested):

src/app/shared/components/product/product-quantity/product-quantity.component.ts
MoritzRS: If the number is negative, it is also below the minOrderQuantity. So negative numbers already trigger a warning. Problem with #849 is, that invalid numbers only show a warning but don't disable the checkout button. This PR was just an idea to solve this issue by not letting the user even enter invalid values. But this solution only prevents invalid characters not invalid quanities. And preventing the user from also entering invalid (positive) quantities would affect the user experience maybe a bit too much. Thinking about that, this whole PR might be kind of pointless as it doesn't really fix anything or bring some kind of actual use. So in the end sorry for wasting time, I guess?

dhhyi: Oh I see, yes. Then it won't be a fix for the issue. But I think adding a pattern and preventing the paste is also a good improvement (and definitely not a waste of time :P).

I think we should take it like that. @shauke agreed?

dhhyi (approved):

src/app/shared/components/product/product-quantity/product-quantity.component.ts
shauke: The changes of this PR can definitely not be a fix for #849. The problem there is that the checkout button is somehow not connected to the invalid state of the form. So preventing invalid input will somewhat help with this issue but not fix it for all relevant problems, e.g. a missing cost center selection. This needs to be fixed elsewhere.
Never the less this PR is not a waste of time since we all learned something and the input pattern definitely improves something. But I removed the input and paste handling. Chromium based browsers will already handle this pretty well on there own and prevent invalid inputs based on the provided patterns and in any other case a validation error message is provided. And if someone really wants to paste invalid content at this place in Firefox or Safari why shouldn't we let them? But it is good to know how we could prevent such things if it would matter. ;)

shauke (approved):

PR #886, opened 2021-09-22
feature: nginx caching param blacklist

MaxKless:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

There is currently no way to modify the cache_key that nginx uses for caching. For example, some meaningless but prevalent query parameters will lead to different cache keys and thus, inefficient caching.

## What Is the New Behavior?

Now, there is a list of parameters that will be ignored during caching. These values can be set to something other than the default by providing a file or list as an environment variable.
Additionally, the nginx folder structure has been modified to allow for cleaner a application of gomplate. 

[ ] Yes
[x] No


MaxKless:
ToDo / Consider:
- [ ] enable nginx JIT compliation for faster regexes
- [ ] find a better solution alltogether?
- [ ] modify the default list of blacklisted parameters?

nginx/caching-ignore-params.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - utm_id
+  - gclid
+  - gclsrc
+  - dclid
+  - fbclid
+  - gbraid
+  - wbraid
+  - _gac
+  - _gcl
+  - epik
```
dhhyi, line 17: should there also be SSO code flow query params like `code`, `state` and `session_state` in this list?

nginx/templates/cache-blacklist.conf.tmpl
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  set $c_uri $is_args$c_uri;
+
+  # remove lonely ?
+  if ($c_uri ~ ^\?$) {
+      set $c_uri "";
+  }
+
+  # finally we have stripped out parameters and have nice cache key
```
dhhyi, line 19: Did you get inspired by a blog post or stackoverflow question for this? If yes, It might be good to add it here as a comment. 😊

nginx/Dockerfile
```diff
@@ -8,8 +8,8 @@ FROM scratch as configstep
 COPY --from=nginx:mainline /etc/nginx /etc/nginx
 COPY nginx.conf /etc/nginx/
 COPY features /etc/nginx/features/
-COPY channel.conf.tmpl /etc/nginx/conf.d/
-COPY entrypoint.sh multi-channel.yaml /
+COPY templates /etc/nginx/templates/
+COPY entrypoint.sh multi-channel.yaml caching-ignore-params.yaml /
```
dhhyi, line 12: using `*.yaml` might make things easier for the future

docker-compose.yml
```diff
@@ -69,3 +69,5 @@ services:
           - baseHref: /b2b
             channel: default
             theme: blue
+      # CACHING_IGNORE_PARAMS: |
```
dhhyi, line 72: please add this example above the `MULTI_CHANNEL`, as it is quite lengthy

nginx/templates/cache-blacklist.conf.tmpl
MaxKless: Yes, heavily _inspired_.

dhhyi (approved):
Looks/works good to/for me 👍

Eisie96 (approved):

PR #890, opened 2021-09-29
perf: refactor loading of server-side translations

dhhyi:
## PR Type

When using a lot of server-side translations like https://pwa-ish-demo.test.intershop.com/INTERSHOP/rest/WFS/inSPIRED-inTRONICS-Site/rest;loc=en_US/localizations?searchKeys=pwa-, the SSR rendering process has performance issues for every rendering.

## What Is the New Behavior?

- Refactored `filterAndTransformKeys` function of `src/app/core/services/localizations/localizations.service.ts` to be mutable, but performant.

[ ] Yes
[x] No


src/app/core/utils/translate/pwa-translate-compiler.ts
```diff
@@ -126,29 +126,28 @@ export class PWATranslateCompiler implements TranslateCompiler {
...
-      })
-      .reduce<Translations>((acc, [key, value]) => {
-        acc[key] = value;
-        return acc;
-      }, {});
+    // This implementation is mutable by intention
+    // tslint:disable-next-line: forin - object does not have inherited properties
+    for (const key in translations) {
+      const value = translations[key];
+      if (this.sanityCheck(value) && this.checkIfCompileNeeded(value)) {
```
MaxKless, line 147: Why do we run the sanity check twice? Once here and once in the compile method.

src/app/core/utils/translate/icm-translate-loader.ts
```diff
@@ -1,47 +1,40 @@
...
-        combineLatest([local$, server$]).pipe(
-          map(([localTranslations, serverTranslations]) => ({
-            ...localTranslations,
-            ...serverTranslations,
-          }))
-        )
+    const server$ = iif(
+      () => isPlatformBrowser(this.platformId) && this.transferState.hasKey(SSR_TRANSLATIONS),
+      of(this.transferState.get(SSR_TRANSLATIONS, {})),
+      this.localizations.getServerTranslations(lang).pipe(
```
MaxKless, line 27: Now that the loader uses the localizations service directly, the ´loadServerTranslations´ actions have become obsolete. We should remove them and their accompanying effect & reducer logic.

MaxKless (changes requested):
Looks pretty good except for a few small things.
Something else I just noticed:
In the fallback translation handler, shouldn't we only log a missing translation warning if the key is not an error key. Because for error keys it is to be expected that the translation will be missing?

src/app/core/utils/translate/pwa-translate-compiler.ts
dhhyi: Good point. As `compileTranslations` is reusing `compile`, we can skip the checks and directly reuse.

src/app/core/utils/translate/icm-translate-loader.ts
dhhyi: 👍

dhhyi:
> In the fallback translation handler, shouldn't we only log a missing translation warning if the key is not an error key. Because for error keys it is to be expected that the translation will be missing?

I wouldn't say it is expected. Especially if a project is not using the ICM localization management as their source for localizations, the log messages help to gather all keys, that need to be translated.

MaxKless (approved):

PR #894, opened 2021-09-30
Fix/upgrade pwa pin versions

dhhyi:
## PR Type

[x] Other: Dependency Upgrade

## What Is the New Behavior?

- `upgrade-pwa.js` script pins down versions correctly
- preparation for next upgrade

[ ] Yes
[x] No


scripts/upgrade-pwa.js
```diff
@@ -17,34 +21,75 @@ const parseVersion = version => {
...
+    } else {
+      packageJson.devDependencies[dep] = write;
+    }
+  });
+  return packageJson;
+});
+
+// install pinned versions
+execute('npm i');
+commit('chore: install pinned versions');
```
shauke, line 92: Why is this step needed that temporary installs older versions of the pinned dependencies?

image

dhhyi: It doesn't install "older" dependencies. `^4.0.0` means for example: take the newest major 4 version. `~4.0.0` would mean: take the newest 4.0 minor version, and so on.
See also: https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

In this temporary step, the right pinned versions are just set in the package.json, so they are respected in the next step.
Later on there is a step that puts in the actual versions.

shauke (approved):

PR #898, opened 2021-10-04
refactor: remove b64u library and use native methods

dhhyi:
## PR Type

[`b64u`](https://www.npmjs.com/package/b64u) library is used for base64 encoding and decoding. This was primarily used for handling the old ICM filter API responses on SSR, as there are no native methods available.

## What Is the New Behavior?

- `b64u` is removed
- remaining implementations (logging in, card payment) are refactored to use the native methods, as they don't happen on SSR

[ ] Yes
[x] No

## Other Information

If base64 functionality is again required on SSR, the use of a polyfill like [btoa-polyfill](https://www.npmjs.com/package/@kfiros/btoa-polyfill) should be preferred.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5400,11 +5400,6 @@
         "ast-types-flow": "0.0.7"
       }
     },
-    "b64u": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/b64u/-/b64u-3.0.0.tgz",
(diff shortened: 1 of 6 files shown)


shauke (approved):

shauke:
We should add a "Breaking Changes" note when merging!

> BREAKING CHANGE:  The PWA no longer relies on the dependency to `b64u` since no encoding and decoding Base64 is needed any more in SSR. In the browser native methods are used. If your custom code uses `b64u` in SSR you should not remove this dependency or use a polyfill like `btoa-polyfill`.

PR #899, opened 2021-10-04
feat: add option to enable SSR only for crawlers in nginx

dhhyi:
## PR Type

[x] Feature
[x] Other: Performance for high-traffic websites

SSR renders are always executed, even for browsers, that could well live without a prerendered page.

## What Is the New Behavior?

Added `SSR` environment variable to nginx to disable SSR rendering for browsers. Crawlers will still get a SEO-complete SSR response.

[ ] Yes
[x] No

## Other Information

[AB#70257](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/70257)


nginx/templates/multi-channel.conf.tmpl
```diff
@@ -46,6 +49,8 @@
 
         rewrite '^(.*)$' '$1$default_rewrite_params' break;
 
+        include /etc/nginx/conf.d/cache-blacklist.conf;
```
MaxKless, line 52: I get why this way of structuring the blacklisting features is cleaner and more contained but I think it's a bad idea to include the cache-blacklist.conf in the location template as I would rather avoid random, seemingly untraceable errors:
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

If you know that this is safe to do in this case, that's something different, of course.

scripts/check-file-synchronization.js
```diff
@@ -1,6 +1,4 @@
-const synchronizations = {
-  'pwa-docker-build': ['Dockerfile', 'Dockerfile_noSSR'],
-};
+const synchronizations = {};
```
MaxKless, line 1: Now that the synchronizations object is empty, this script doesn't do anything anymore, does it?
Are you keeping it for those projects who do need it? Or can we remove it?

nginx/templates/cache-blacklist.conf.tmpl
```diff
@@ -1,5 +1,7 @@
 # inspired by https://gist.github.com/a-vasyliev/de8ffc6c6aa74cdeadfe
 
+set $c_uri $args;
```
MaxKless, line 3: Good point, I don't know why I didn't put in in here in the first place^^

nginx/templates/multi-channel.conf.tmpl
```diff
@@ -28,13 +28,16 @@
...
         proxy_cache_valid 200 302 {{ getenv "CACHE_DURATION_NGINX_OK" }};
         proxy_cache_valid 404     {{ getenv "CACHE_DURATION_NGINX_NF" }};
-        # set $c_uri as cache_key
-        proxy_cache_key $c_uri;
 
         add_header X-icm-channel {{ $channel }} always;
         add_header X-icm-default-lang {{ $lang }} always;
         add_header X-ua-device $ua_device always;
 
+        if ($render_ssr = 0) {
```
MaxKless, line 36: Same here, I'm not sure what the consequences of putting the if block into the location template would be. The SSR rendering doesn't have to be configurable per baseHref, does it? So maybe we can move it outside.

MaxKless (changes requested):

nginx/templates/multi-channel.conf.tmpl
dhhyi: It looks like it doesn't work outside of this location block. Does it work for you?

nginx/templates/multi-channel.conf.tmpl
dhhyi: If I put it outside, the cache key calculation is not done correctly and URLs with query params do not get mapped to `/loading` without query param.

nginx/templates/cache-blacklist.conf.tmpl
dhhyi: 👍

jometzner: I had to search the internet for the meaning of `$args` and found that it is esentialy the same as `$query_string` which I find more apropriate. Dunno what you guys think.

https://nginx.org/en/docs/http/ngx_http_core_module.html#var_args

MaxKless:
Postponing this PR for now - will come back to it later.

MaxKless (approved):

PR #902, opened 2021-10-07
fix: wait for retrieving localized ICM data until configuration is retrieved

dhhyi:
## PR Type

[x] Bugfix

Locale information is not correctly applied before localized data is retrieved.
Steps:
- go to: https://intershoppwa.azurewebsites.net/catComputers.191;lang=de_DE
- category data is still in English

## What Is the New Behavior?

Fixed by refactoring and prioritizing `configurations` call.

- ApiService property `runExclusively` was not working anymore and has become obsolete with identity providers.
- Replaced by `sendLocale` and `sendCurrency` properties (defaulting to true), that wait for the correct locale info to be set.

[ ] Yes
[x] No


src/app/core/store/core/configuration/configuration.selectors.ts
```diff
@@ -41,10 +41,8 @@ const internalCurrencyFilter = createSelector(getConfigurationState, state => st
  *
  * if no locale is available, then a default configured locale from environment.ts is loaded as fallback
```
MaxKless, line 42: This comment doesn't seem accurate anymore.

MaxKless (changes requested):
Minor change - only a comment.

PR #907, opened 2021-10-18
fix: check for whole occurrences of localizations

dhhyi:
## PR Type

[x] Bugfix

`clean-localizations` script searches for occurrences of keys in source files, but does not check it the key is part of a different key.

Example: key `account.organization.user_management.user_detail` is kept, because key `account.organization.user_management.user_detail.breadcrumb` is still used.

## What Is the New Behavior?

Whole occurrence is checked with word-boundary matching.

[ ] Yes
[ ] No

## Other Information

[AB#70571](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/70571)


dhhyi:
@suschneider I just had another thought, the dots in the key have to be escaped and maybe dashes should also be disallowed as word boundaries. I'll add something tomorrow.

PR #911, opened 2021-10-22
feat: store locator

### LINKED ISSUE #754: Store locator missing on PWA

On the PWA version I didn't find how to reach the store locator search engine.

Having the search engine for store locator like on classic version of Intershop

no alternatives identified

shauke:
## PR Type

[x] Feature

Currently the PWA does not provide Store Finder functionality similar to the one in the Responsive Starter Store
https://pwa-review.northeurope.cloudapp.azure.com/INTERSHOP/web/WFS/inSPIRED-inTRONICS_Business-Site/en_US/-/USD/ViewStoreFinder-Start

Issue Number: Closes #754

## What Is the New Behavior?

Store Finder functionality is available and configurable via feature toogle.
In addition Google Maps integration can be activated.

Store Finder with Google Maps integration is activated by default for the B2B theme
http://pwa-gh-review-911.azurewebsites.net/store-finder

Store Finder without Google Maps integration is the default configuration for the B2C theme
http://pwa-gh-review-911.azurewebsites.net/store-finder;theme=b2c

[x] No

## Other Information

[AB#70748](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/70748)

## To Dos

- [ ] IAD/VD review
- [x] Store Finder or Store Locator link integration
- [x] code review
- [x] documentation
- [x] waiting for REST API ["'stores' resource should return custom attributes 'Latitude' and 'Longitude'"](https://dev.azure.com/intershop-com/Products/_workitems/edit/70652)
- [x] remove mock data and docker ignore adaption
- [x] remove maps key from env
- [x] scroll behavior parent
- [x] rebase for git credit
- [x] handle "no results" response
- [x] update linting to ESLint
- [x] api key as env var
- [x] ability to provide highlight icon
- [x] fix tests


M-Behr (changes requested):
I like the google maps integration 👍 Nice feature!

We would like to change a little bit the user experience design. Here are the IAD/VD remarks:
- [x] Small text changes in Store finder introduction (see designs below).
- [x] Space between paragraph and store finder form is too large. I think it should be 15px between the explanation "Simply select a country..." and the form (see registration form as example).
- [x] Typo: ZIP/Postal Code (see labels in registration form - each word starts with capital letters, no blank character after /)
- [x] "Results of store location finder" heading is no longer displayed - only a counter for the results (paragraph like "no-search-result-title").
- [x] Display a list of stores on the left side (with scrollbar) and map is displayed on right side.
- [x] If user hovers over a list item (background-color: #e6e6e6), the pin is highlighted (highlight definition is in progress). Triggered by click on mobile devices.
- [x] Please display the e-mail in the store contact information.
- [x] Use icons for phone, fax and e-mail (fa-phone, fa-fax, fa-envelope).
- [x] Phone and e-mail are links (tel: and mailto:) so that users can contact the store immediately without copying the number or e-mail address.
- [x] Phone view: Display store list below map (see designs below).

Desktop:
Map
Phone:
Phone-Map
No map:
List
No results:
Empty

iwiederhold (dismissed):
Great work! THX

M-Behr (changes requested):
It looks fantastic. I would like to change one last things:

- [x] Use the default margin of 16px bottom for paragraph text. I think the PWA uses only "p" (see contact form) or "p class=ng-star-inserted" (see my account pages) most of the cases. Your paragraph uses class="my-4" with 24px margin top and bottom (see screenshot below). @MoritzRS: Why you use this class? Is it something special?
- [x] Please use the solid envelope as icon for e-mail (i class="fas fa-envelope"). I used the false icon for the design. Sorry, this was my mistake.
- [x] If you click on a pin on the map, then the list should move to the selected item on the left side.
- [ ] Please change the selection of a list item from hover to click on desktop view. During the review, the behavior feels unsteady and the user losses control (example: I scroll through the list and then the selected pin is changed without me wanting it). It feels much better on the current behavior in mobile view. That's why I decided to make this change.

2021-11-09 11_33_40-Intershop Progressive Web App _ Intershop PWA

MoritzRS:
Scrolling itself should be supported on all modern browsers (uses Element.scrollTo). However smoothing via option.behavior might not be supported on every browser. Works fine on Chrome and Firefox but couldn't use smooth on Epiphany (WebKit based) so Safari might also not be working as well. Depending on the use case, polyfills might be needed.

M-Behr (changes requested):
It looks and works very well.
Currently the list items look not clickable.
2021-11-10 09_26_53-Intershop Progressive Web App _ Intershop PWA

Therefore we have to add 2 little things:
- [x] The hand as mouse pointer, which indicates an action. You can see it when you hover over buttons, links or pins.
- [x] And a hover styling is needed (background color e6e6e6).

These are the last changes, I promise! :)

M-Behr (approved):
Great job 👍

src/app/extensions/store-locator/models/store/store.mapper.ts
```diff
@@ -0,0 +1,28 @@
...
+        name: storeData.name,
+        address: storeData.address,
+        city: storeData.city,
+        postalCode: storeData.postalCode,
+        country: storeData.country,
+        countryCode: storeData.countryCode,
+        email: storeData.email,
+        fax: storeData.fax,
+        phone: storeData.phoneBusiness,
+        latitude: Number(storeData.latitude),
```
MaxKless, line 21: Not saying that it isn't, but are you sure using `Number` for the casting is the right tool for conversion here? 
We wouldn't want leading zeroes or any weird behavior to become a problem: [Casting Methods](https://stackoverflow.com/a/14668510/16965087)

src/app/extensions/store-locator/pages/store-locator/store-locator-page.component.ts
```diff
@@ -0,0 +1,118 @@
...
+        key: 'countryCode',
+        type: 'ish-select-field',
+        wrappers: ['form-field-horizontal'],
+        templateOptions: {
+          required: false,
+          label: 'store_locator.form.country.label',
+          placeholder: 'store_locator.form.country.default',
+          options: this.appFacade
+            .countries$()
+            ?.pipe(
```
MaxKless, line 51: `appFacade.countries$()` always returns an observable, and even if that doesn't contain any values, it should never be falsy. Instead, you should probably move the optional chaining operator to `countries.map`, because the first emission could very well be falsy.

src/app/extensions/store-locator/pages/store-locator/store-locator-page.component.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  empty(stores: StoreModel[]) {
+    return !stores?.length;
+  }
+
+  highlighted(store: StoreModel, highlight: StoreModel) {
+    return StoreHelper.equal(store, highlight);
+  }
+
+  scrollTo(container: HTMLElement, element: HTMLElement) {
```
MaxKless, line 93: Sorry if this is a dumb question, I'm not 100% sure about the setup here, but have you made sure that this function can never be called by the SSR container? Using `window` doesn't work in SSR so sometimes you have to guard these function calls with `isPlatformBrowser`.

src/app/extensions/store-locator/models/store/store.mapper.ts
MoritzRS: `Number` shouldn't throw any errors, however `parseFloat` would be more specific as coordinates are pairs of floats. I think I'll change it, so it's at least clear what kind of number we're talking about

src/app/extensions/store-locator/pages/store-locator/store-locator-page.component.ts
MoritzRS: `scrollTo` gets called the first time after user searches for stores and stores were loaded (after `loadStores` action gets dispatched). So it shouldn't get called during SSR and only after user actually interacts with the page

src/app/extensions/store-locator/pages/store-locator/store-locator-page.component.ts
MoritzRS: I don't even know, why I put it there in the first place. I think it was because I ran into some kind of error but I couldn't replicate any of it with the current version. Think this can safely be removed now

src/app/extensions/store-locator/pages/store-locator/store-locator-page.component.ts
MaxKless: Sounds reasonable. Do you think there might be a better location to put the scrolling logic? It isn't really component-specific after all and very reusable...I was thinking of creating a scrollService or at least some static helper methods in `src/app/core/utils`.
Another question: Why don't you just use `element.scrollIntoView()` to control the scrolling instead of the complicated, manual way? Is it that important to be able to control the speed and easing granularly. `scrollIntoView` also supports easing :)

MoritzRS: Extracting it to a service sounds reasonable.

As to your second question, other than solving the smoothing in safari, this custom method also doesn't scroll the whole page down but only the specified scrollable container. I found if somewhat disturbing if the page itself scrolls which is why I wanted to keep it only to the container. Other than that, `scrollIntoView` with smoothing creates a weird bug when clicking on search more than once, where the list scrolls all the way to the end for no reason. The custom method was created to deal with that problem in the first place.

MaxKless: Makes sense. The scrolling was more out of curiosity, the service extraction is what I think is important :D So please do that and good luck with the scrolling solution 👍

M-Behr (approved):

PR #923, opened 2021-11-08
refactor: rename themes - (b2b + b2c)

shauke:
## PR Type

[x] Other: rename standard themes

The Intershop PWA standard themes are currently named `default` (for B2C theme) and `blue` (for B2B theme). Unfortunately does this naming not really fit good together. It is also not to fortunate that the B2C theme is called `default` and used as default while the Intershop PWA focus lies on the B2B functionality.

## What Is the New Behavior?

The standard themes are renamed to `b2b` (was `blue`) and `b2c` (was `default`) and the `b2b` theme is configured as the default theme.

[x] Yes

BREAKING CHANGES: renaming of theme related files and folders

## Other Information

[AB#71157](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/71157)


angular.json
```diff
@@ -32,12 +32,12 @@
             "aot": true,
             "assets": ["src/assets"],
             "scripts": [],
-            "styles": ["src/styles/themes/default/style.scss"],
+            "styles": ["src/styles/themes/b2b/style.scss"],
             "stylePreprocessorOptions": {
-              "includePaths": ["src/styles/themes/default"]
+              "includePaths": ["src/styles/themes/b2b"]
```
dhhyi, line 37: I would honestly leave this at "default" or rename it "placeholder" to showcase that is will be replaced by the build system. 🤔

Same for the other default paths. Changing the`defaultConfiguration`s is instead mandatory.

nginx/README.md
```diff
@@ -9,19 +9,19 @@ Please refer to the [documentation](../docs/guides/nginx-startup.md) for configu
...
      lang: en_US
+     theme: 'b2c'
    .+\.de:
      channel: inSPIRED-inTRONICS-Site
      lang: de_DE
+     theme: 'b2c'
    .+\.com:
      channel: inSPIRED-inTRONICS_Business-Site
      features: quoting,businessCustomerRegistration,advancedVariationHandling
-     theme: 'blue|688dc3'
```
dhhyi, line 18: This change is overdue. Color codes should current no longer work 😅

package.json
```diff
@@ -81,7 +81,7 @@
     "angular-oauth2-oidc": "^12.1.0",
     "angular2-uuid": "^1.1.1",
     "angulartics2": "^10.0.0",
-    "bootstrap": "^4.6.0",
+    "bootstrap": "^4.6.1",
```
dhhyi, line 84: Unrelated change

dhhyi (changes requested):

angular.json
shauke: Changed the `stylePreprocessorOptions` to use `placeholder` instead of `b2b`.

shauke: @dhhyi did I find the right places?

templates/webpack/webpack.custom.ts
```diff
@@ -296,8 +296,8 @@ export default (config: Configuration, angularJsonConfig: CustomWebpackBrowserSc
     });
   });
 
-  if (theme !== 'default') {
-    const defaultThemePath = join('src', 'styles', 'themes', 'default');
+  if (theme !== 'b2b') {
+    const defaultThemePath = join('src', 'styles', 'themes', 'b2b');
```
dhhyi, line 300: this `if` must now be removed and the `'b2b'` must become `'placeholder'`, otherwise the component style imports won't work.

You can test that by adding a component style override with `@import 'variables';`.
I think this test can also be added to the `test-schematics-customization.sh`

angular.json
```diff
@@ -32,12 +32,12 @@
             "aot": true,
             "assets": ["src/assets"],
             "scripts": [],
-            "styles": ["src/styles/themes/default/style.scss"],
+            "styles": ["src/styles/themes/b2b/style.scss"],
```
dhhyi, line 35: This must also become `placeholder` to be properly replaced. For this you also have to create the file to satisfy Angular configuration validation.

dhhyi: Let me try something 😃

dhhyi (changes requested):

angular.json
shauke: I had changed this too but then the build results in an error.

templates/webpack/webpack.custom.ts
dhhyi: see 61b006e

angular.json
dhhyi: Angular config validation expects this file to exist and it is run before the webpack.custom replaces it. So as long as the replacing is not done by an Angular builder, it has to be that way. 😅 

see 61b006e

dhhyi (approved):

SGrueber (changes requested):

docs/guides/multi-site-configurations.md
```diff
@@ -180,24 +180,23 @@ To see what is possible through multi-site handling, have a look at this extende
...
 .+\.de:
   channel: inSPIRED-inTRONICS-Site
   lang: de_DE
+  theme: b2c
 .+\.com:
   channel: inSPIRED-inTRONICS_Business-Site
   features: quoting,businessCustomerRegistration,advancedVariationHandling
-  theme: blue
 .+\.fr:
   channel: inSPIRED-inTRONICS-Site
```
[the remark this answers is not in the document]
shauke: fixed the documentation inconsistency

SGrueber (approved):

PR #927, opened 2021-11-15
feat: allow translations to be loaded from an external source

dhhyi:
## PR Type

[x] Feature

Translation cannot be loaded from an external source.
ICM translation management has some limitations:
- Translations can't be shared between locales of brands (depending on the channel setup)

## What Is the New Behavior?

An external translation source can be defined.

### For testing:

1. start external translation source: `npx http-server src/assets/i18n --cors`
2. set `translateLoaderUrl: 'http://localhost:8080'` in `environment.development.ts`
3. play around with `translateLoaderFormat: '$language_$COUNTRY.json'` in `environment.development.ts`

[ ] Yes
[ ] No

## Other Information

[AB#71316](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/71316)

### DIFF
--- a/docs/concepts/localization.md
+++ b/docs/concepts/localization.md
@@ -260,6 +260,12 @@ This has to fit to your PWA environment.
 This allows to enable easy filtering.
 The prefixes will be removed when the translations are loaded into the PWA.
 
+## External Localization Sources
+
+The Intershop PWA can be configured to use an external source for loading translations in JSON format.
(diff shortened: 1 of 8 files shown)


shauke:
We are interested in feedback in how many projects would want to use an external translation system instead of the localization functionality of ICM used by default.

shauke:
Thank you @dhhyi for this pull request. But for limited request from other projects we will mot merge it into the standard Intershop PWA project and it needs to stay a project customization for you. Sorry for the late response.
We will add a link to this closed PR in the `localization.md` for projects that might be interested in your idea in the future.

PR #930, opened 2021-11-22
fix: allow + sign in email addresses

DiverDori:
## PR Type

[x] Bugfix

Currently, email addresses like `diverdori+6@<randomemailaddressprovider>.com` are not supported.
`+` is a common sign in email.

## What Is the New Behavior?
With this fix email addresses like `diverdori+6@<randomemailaddressprovider>.com` are supported.

[ ] Yes
[x] No

## Other Information

[AB#71504](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/71504)

### DIFF
--- a/src/app/shared/forms/validators/special-validators.ts
+++ b/src/app/shared/forms/validators/special-validators.ts
@@ -57,7 +57,7 @@ export class SpecialValidators {
      * - no IPs allowed for login emails
      * - only some special characters allowed
      */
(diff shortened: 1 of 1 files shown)


shauke:
Allowing `+` signs in email addresses that are used as login id as well has side impacts tat need to be handled as well, e.g. double encoding of user ids in REST requests. There is still discussion and work to do for this fix.

shauke:
Fixed with #940

PR #940, opened 2021-12-01
fix: use double encoded resource ids for user login

### LINKED ISSUE #939: Special Characters in email-address lead to problems with server communication when only encoded once
**Actual Behavior**
When a registered b2b user with special characters in email address, like "diver+dori@<someEmailProvider>.com" tries to use the PWA, there are problems in the server communication. 

**Expected Behavior**
Everything should work. 

1. Try to somehow register a user with special character, this can also happen by changing the email address of an existing user
2. surf in the personal section with this user in the PWA
3. --> a lot of 4xx in server communication / a lot of missing data

**Environment Details**
any environment

There are ICM server side fixes in 7.10.38 and 7.10.26.

DiverDori:
## PR Type
[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

When a registered b2b user with special characters in email address, like "diver+dori@.com" tries to use the PWA, there are problems in the server communication.

Issue Number: Closes #939

## What Is the New Behavior?
Everything works.

[ ] Yes
[x] No


src/app/core/utils/url-resource-ids.ts
```diff
@@ -0,0 +1,3 @@
+export function encodeResourceID(object: string): string {
+  return encodeURIComponent(encodeURIComponent(object));
+}
```
shauke, line 3: This would be the place to write a comment that `encodeResourceID` currently means "double encoding" and a short explanation why?

shauke (changes requested):

src/app/core/utils/url-resource-ids.ts
dhhyi: I would actually implement this as a HttpInterceptor, to make sure it is used everywhere.

dhhyi: I really think this is an ICM issue and not a PWA issue 🤔

DiverDori: We had a very close look into this, and with the email addresses being used currently in the requests, it cannot be solved within the ICM. There is a team looking into using other IDs (without special characters), but this will take time and we will still need a fix for the email addresses for legacy reasons. I will look into HttpInterceptors though.

dhhyi: You can also introduce an interceptor that always does `return next.handle(req.clone({ url: encodeURI(req.url) }));`. Works also for other cases. 😉

DiverDori: The http interceptor seemed very promising. Unfortunatelly this does not work. It either encodes not enough (encodeURI does not encode a lot of characters that we need encoded) or not enough (encodeURIComponent does encode everything if called with the whole URI). 
Therefore it still, sadly, seems like the best solution to directly encode the specific resource ids.

dhhyi: @DiverDori So as far as I can see for now you only want to double encode the email part in the URL, correct? This can still be done with an interceptor. Let me try.

dhhyi:
```
export class EncodeEmailInterceptor implements HttpInterceptor {
  intercept(req: HttpRequest<unknown>, next: HttpHandler) {
    if (decodeURIComponent(req.url).includes('@') && !req.url.includes('/quotes/')) {
      const regex = /^.*\/(.*?@.*?)\/.*$/;
      const match = regex.exec(decodeURIComponent(req.url));

      const email = match?.[1];
      if (email) {
        const newURL = decodeURIComponent(req.url).replace(email, encodeURIComponent(encodeURIComponent(email)));
        return next.handle(req.clone({ url: newURL }));
      }
    }
    return next.handle(req);
  }
}
```
That would be just replacing the email. But I can see that quoting is no longer working for standard users. So is it even correct that you double-encoded the quoting related services in this PR?

DiverDori: If this was in Java, I would decline. This is effort (decode and includes()) in every case, without being able to handle everything (like starting to include customerID to be encoded twice). 
The fix for quotes and quoterequests is in ICM and not yet released. I have to talk to the PWA team how this is usually handled.

dhhyi: Yes, this should be restricted to `PLATFORM_SERVER` for sure, you're right! But apart from that I don't see an issue delegating this extra effort to all client applications. 😃

SGrueber:
Many e2e tests fail, because this PR requires an ICM 7.10.38. Against lower ICM versions many requests respond with a 403 error, e.g.

http://pwa-review.northeurope.cloudapp.azure.com:8081/INTERSHOP/rest/WFS/inSPIRED-inTRONICS_Business-Site/rest;loc=en_US;cur=USD/customers/OilCorp/users/bboldner%2540test.intershop.de/costcenters
http://pwa-review.northeurope.cloudapp.azure.com:8081/INTERSHOP/rest/WFS/inSPIRED-inTRONICS_Business-Site/rest;loc=en_US;cur=USD/customers/OilCorp/users/bboldner%2540test.intershop.de/requisitions?view=buyer&status=PENDING
http://pwa-review.northeurope.cloudapp.azure.com:8081/INTERSHOP/rest/WFS/inSPIRED-inTRONICS_Business-Site/rest;loc=en_US;cur=USD/customers/OilCorp/users/bboldner%2540test.intershop.de/quoterequests?attrs=number,name,lineItems,creationDate,submittedDate
http://pwa-review.northeurope.cloudapp.azure.com:8081/INTERSHOP/rest/WFS/inSPIRED-inTRONICS_Business-Site/rest;loc=en_US;cur=USD/customers/OilCorp/users/bboldner%2540test.intershop.de/quotes?attrs=number,name,lineItems,creationDate,validFromDate,validToDate,rejected

src/app/core/utils/url-resource-ids.ts
shauke: We will use the `encodeResourceID` method for the places needed.
If a project still needs the former way of handling the encoding it is just this one place in this method where the project can replace the double encoding with a single encoding.

shauke (dismissed):

shauke (approved):

PR #949, opened 2021-12-07
Scrolling to error message

### LINKED ISSUE #848: Automatic scroll up to inline error messages

I'm always frustrated when "Continue Checkout" is disabled and I don't see the error message at first glance (see screenshot).

Inline messages are displayed below the headline. The checkout button is usually further down on the page (especially in mobile). If a user clicks "Continue Checkout" and an error is displayed, the button is disabled. However, the error is not immediately visible. You have to scroll up first.

If an error message is displayed, scroll automatically to the message.

If an error message is displayed, scroll automatically to top of the page. Inline messages are below page title. They should then also be in the visible area.

2021-08-25 09_40_20-Checkout _ Intershop PWA

[AB#69214](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/69214)

MoritzRS:
## PR Type

[x] Feature

On checkout error message doesn't get scrolled to
Issue Number: Closes #848 

## What Is the New Behavior?
When the checkout has some errors, the browser will scroll to it

[x] No

## Other Information
Uses Scrolling Directive developed for #911 

[AB#71917](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/71917)

### DIFF
--- a/src/app/core/directives/scroll.directive.ts
+++ b/src/app/core/directives/scroll.directive.ts
@@ -1,0 +1,120 @@
+import { Directive, ElementRef, Input, OnChanges } from '@angular/core';
+
+/**
+ * Structural directive.
+ * Used on an element, the elements parent will scroll to it, when the given value is true
+ *
+ * @example
+ * <div [ishScroll]="true" [scrollDuration]="500">
+ *   Parent will scroll smoothly to this element within 500 milliseconds
+ * </div>
+ * or
+ * <div [ishScroll]="true">
+ *   Parent will scroll to this element instantly
+ * </div>
+ * or
+ * <div [ishScroll]="false">
+ *   Parent will not scroll to this element
+ * </div>
+ */
+@Directive({
+  selector: '[ishScroll]',
+})
+export class ScrollDirective implements OnChanges {
+  constructor(private el: ElementRef) {}
+
+  /**
+   * Wether or not scrolling should happen
+   */
+  @Input() ishScroll: boolean;
+
+  /**
+   * Sets the duration for the scrolling in milliseconds.
+   * If set to 0 scrolling happens instantly.
+   */
+  @Input() scrollDuration = 0;
+
+  /**
+   * Sets the scroll container
+   * This values uses the elements parent node by default.
+   *
+   * @usageNotes
+   * Set it to "root" to use the window, "parent" to use the elements parent (default)
+   * or pass in any HTMLElement that is a parent to the element.
+   */
+  @Input() scrollContainer: 'parent' | 'root' | HTMLElement = 'root';
+
+  /**
+   * Sets spacing above the element in pixel.
+   * If set, scrolling will target given amount of pixel above the element.
+   * Set to 0 by default
+   */
+  @Input() scrollSpacing = 0;
+
+  ngOnChanges() {
+    if (this.ishScroll) {
+      this.scroll();
+    }
+  }
+
+  private scroll() {
+    const target: HTMLElement = this.el.nativeElement;
+    const scrollContainer =
+      this.scrollContainer === 'parent'
+        ? target.parentElement
+        : this.scrollContainer === 'root'
+        ? document.documentElement
+        : this.scrollContainer;
+
+    // calculate the offset from target to scrollContainer
+    let offset = target.offsetTop;
+    let tempTarget = target.offsetParent as HTMLElement;
+    while (!tempTarget.isSameNode(scrollContainer) && !tempTarget.isSameNode(document.body)) {
+      offset += tempTarget.offsetTop;
+      tempTarget = tempTarget.offsetParent as HTMLElement;
+    }
+    offset -= this.scrollSpacing;
+
+    // scroll instantly if no duration is set
+    if (!this.scrollDuration) {
+      scrollContainer.scrollTop = offset;
+      return;
+    }
+
+    // set values for animation
+    let startTime: number;
+    const initialScrollPosition = scrollContainer.scrollTop;
+    const scrollDifference = offset - initialScrollPosition;
+
+    const step = (timestamp: number) => {
+      // set start time at the beginning
+      if (!startTime) {
+        startTime = timestamp;
+      }
+
+      // get time passed
+      const passed = timestamp - startTime;
+
+      // exit animation when duration is reached or calculated difference is smaller than 0.5 pixel
+      if (passed >= this.scrollDuration || Math.abs(scrollContainer.scrollTop - offset) <= 0.5) {
+        scrollContainer.scrollTop = offset;
+        return;
+      }
+
+      // current progress of the animation based on timing
+      const progress = passed / this.scrollDuration;
+
+      // cubic ease-in-out function
+      const ease = (x: number) => (x < 0.5 ? 4 * x * x * x : 1 - Math.pow(-2 * x + 2, 3) / 2);
+
+      // set new scroll value based on current time progression
+      scrollContainer.scrollTop = initialScrollPosition + scrollDifference * ease(progress);
+
+      // request next animation frame
+      requestAnimationFrame(step);
+    };
+
+    // start animation by requesting animation frame
+    requestAnimationFrame(step);
+  }
+}
--- a/src/app/core/directives.module.ts
+++ b/src/app/core/directives.module.ts
@@ -5,6 +5,7 @@ import { IdentityProviderCapabilityDirective } from './directives/identity-provi
 import { IntersectionObserverDirective } from './directives/intersection-observer.directive';
 import { ProductContextAccessDirective } from './directives/product-context-access.directive';
 import { ProductContextDirective } from './directives/product-context.directive';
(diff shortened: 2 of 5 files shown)


MoritzRS:
This adds automatic scrolling in checkout step 3 (payment). Please comment if it is required elsewhere.

M-Behr:
> This adds automatic scrolling in checkout step 3 (payment). Please comment if it is required elsewhere.

Hi, it works very well.
I think we should add this behavior to the inline error message in general. This ensures consistent behavior no matter where these messages are used. In addition, automatic scrolling is necessary in the shopping cart and at each checkout step, where there is always a lot of content displayed and the "Checkout" button can be seen far below (especially in the phone view) - This makes it very difficult to notice the error message as I mentioned in the ticket.

MoritzRS:
Alright, the way this current implementation works is by checking the basket error status. I will work on tying it to the inline error message in general so it can be used everywhere.

src/app/pages/checkout-payment/checkout-payment/checkout-payment.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="row" *ngIf="basket">
-  <div class="col-md-12">
+  <div class="col-md-12" [ishScroll]="basketHasError" [scrollContainer]="documentElement" [scrollDuration]="500">
```
MaxKless, line 2: I think the current way you access the `documentElement` is not very pretty. I think it would be a lot cleaner if the component didn't have to know about any DOM logic but instead only takes care of providing data and form logic. My suggestions:
-  Get the `documentElement` in the directive from the target element by going up the tree and finding the root element. You can control this behavior with an Input flag (for example, `@Input() scrollContainer: 'parent' | 'root' | HTMLElement`) and then decide in the directive based on that.
- Definitely don't require the component to get some part of the DOM. It's quite "un-angular" and there are better ways. For example, if you want to provide a specific `Element` that should scroll, you can use a [template reference variable](https://angular.io/guide/template-reference-variables#template-variable-scope) like this: 

```html
<div class="row" *ngIf="basket" #example>
  <div class="col-md-12" [ishScroll]="basketHasError" [scrollContainer]="example" [scrollDuration]="500">
  </div>
</div>
```

Of course you have to take care to see what kind of type `example` will be (I think `ElementRef`) but this would enable you to skip the component and setup this logic directly in the template.

src/app/core/directives/scroll.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Input() ishScroll: boolean;
+  @Input() scrollDuration = 0;
+  @Input() scrollContainer: HTMLElement;
+  ngOnChanges() {
+    if (this.ishScroll) {
+      this.scroll();
+    }
+  }
+
+  private scroll() {
```
MaxKless, line 34: Some documentation would be nice here. Just a few lines above each block to explain what's happening.

src/app/core/directives/scroll.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+ * </div>
+ * or
+ * <div [ishScroll]="false">
+ *   Parent will not scroll to this element
+ * </div>
+ */
+@Directive({
+  selector: '[ishScroll]',
+})
+export class ScrollDirective implements OnChanges {
```
MaxKless, line 23: This directive is quite math-heavy. I think it's important that it is well understood what goes on here so please add some unit tests in addition to more documentation.

src/app/pages/checkout-payment/checkout-payment/checkout-payment.component.ts
```diff
@@ -86,6 +94,11 @@ export class CheckoutPaymentComponent implements OnInit, OnChanges, OnDestroy {
       const redirect = params.get('redirect');
       this.redirectStatus = redirect;
     });
+
+    this.checkoutFacade.basketValidationResults$.pipe(takeUntil(this.destroy$)).subscribe(validationResults => {
```
MaxKless, line 98: Try not to use subscribe, especially in very preventable cases like this. Instead create an observable stream that emits what you want (true when the basket has an error) and then subscribe to it directly in the template using the `async` pipe.

src/app/pages/checkout-payment/checkout-payment/checkout-payment.component.ts
```diff
@@ -86,6 +94,11 @@ export class CheckoutPaymentComponent implements OnInit, OnChanges, OnDestroy {
...
+      // tslint:disable-next-line: no-boolean-literal-compare
+      this.basketHasError = validationResults.valid === false;
```
MaxKless, line 100: You can probably find a clever solution here that doesn't need to disable a tslint rule. If you're concerned about differentiating between nullish values and false, I recommend checking out the [nullish coalescing operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing).

MaxKless (changes requested):
Nice solution but needs some work.

MoritzRS:
Scrolling is now attached in the basket-validation-results component. It will work for all messages used there but is still not attached on messages in general. Here are some thoughts regarding general attaching to any kind of error messages:

1. Is it really desired to automatically scroll to every error message? Some messages e.g. on form validation show up while user is still filling out forms. Automatically scrolling to those messages will result in an unoptimized user experience.
2. If multiple error messages are shown, all will be scrolled to in order which results in scrolling to the last one. This way the user only notices the last message at first glance. This can be problematic since error messages could as well be triggered by preceeding errors.
3. Depending on where messages are shown it is not desired to scroll them directly but to some point above them since the message will otherwise be hidden on mobile by the navigation bar. The scroll directive supports spacing to leave some space above the element when scrolling but I am not sure if this should be set for all messages in general since some cases might require different spacing or no spacing at all.

Considering the points above I advise to simply scroll to the top when an error occurs (mentioned as alternative solution in #848). This behvior might not be archived on a message component but has to be implemented on the corresponding page. Easiest option would be to scroll automatically to the top after the checkout button gets pressed since even on success, user will land at the top for the next checkout step.

Please reflect on these thoughts and decide how to proceed.

M-Behr:
> Scrolling is now attached in the basket-validation-results component. It will work for all messages used there but is still not attached on messages in general. Here are some thoughts regarding general attaching to any kind of error messages:
> 
> 1. Is it really desired to automatically scroll to every error message? Some messages e.g. on form validation show up while user is still filling out forms. Automatically scrolling to those messages will result in an unoptimized user experience.
> 2. If multiple error messages are shown, all will be scrolled to in order which results in scrolling to the last one. This way the user only notices the last message at first glance. This can be problematic since error messages could as well be triggered by preceeding errors.
> 3. Depending on where messages are shown it is not desired to scroll them directly but to some point above them since the message will otherwise be hidden on mobile by the navigation bar. The scroll directive supports spacing to leave some space above the element when scrolling but I am not sure if this should be set for all messages in general since some cases might require different spacing or no spacing at all.
> 
> Considering the points above I advise to simply scroll to the top when an error occurs (mentioned as alternative solution in #848). This behvior might not be archived on a message component but has to be implemented on the corresponding page. Easiest option would be to scroll automatically to the top after the checkout button gets pressed since even on success, user will land at the top for the next checkout step.
> 
> Please reflect on these thoughts and decide how to proceed.

Thank you for your feedback. I think we have a little misunderstanding: I would only add this behavior to the inline error messages - the bootstrap error alerts. Only this pattern/component:
image
Currently we use the inline error messages or error alerts only in the cart and checkout pages. I think it is fine to attach it on the basket-validation-results component. My assumption: If the basket validation fails, then the pattern shown above is always used, isn't it?
1. Form field validation uses another pattern. No auto-scrolling should be used for this.
2. This is rather rarely the case. Even if it were, you could fix one mistake first and then the other. Currently, both errors would be out of view.
3. As I mentioned I would use this for the error inline messages (not all messages in general), which we are only use for cart and checkout now. I have to check how it feels on the cart page, where we have a sticky navigation at the top.

I will check the error behavior in the whole checkout and when something is unclear or feels unnatural, I will come back to you. Is this fine for you? It seems that the demo server is not loading the pwa, yet (504 Gateway time-out). I try it later again.

MoritzRS:
Thanks for clarifying, I guess I did in fact misunderstood your request. 
I would appreciate it if you could review the checkout process once the demo server is back online and inform me about further changes you wish to happen.

M-Behr (changes requested):
The auto-scrolling behavior works very well.
I noticed that some checkout messages do not scroll up automatically. Perhaps some error or info messages are not part of the basket validation result component, e.g. have no shipping address assigned or credit card is invalid.
The auto-scrolling behavior should be implemented for error-message and info-message component, if toast = false. I discussed this solution with @SGrueber.

MaxKless (approved):

M-Behr (approved):
Good job 👍

PR #954, opened 2021-12-09
refactor: feature toggle

cesarGamaSa:
## PR Type

[ ] Bugfix
[ ] Feature
[X] Refactoring (no functional changes, no API changes)
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Feature toggle directive and overall functionality has an old structure.

## What Is the New Behavior?

Feature toggle uses subscription like other directives (e.g: not role toggle directive).

[ ] Yes
[X] No


dhhyi (approved):

dhhyi (approved):

src/app/core/directives/feature-toggle.directive.ts
```diff
@@ -17,34 +19,59 @@ import { FeatureToggleService } from 'ish-core/utils/feature-toggle/feature-togg
...
     private viewContainer: ViewContainerRef,
-    private featureToggle: FeatureToggleService
-  ) {}
+    private featureToggle: FeatureToggleService,
+    private cdRef: ChangeDetectorRef
+  ) {
+    combineLatest([
+      this.enabled$.pipe(
+        distinctUntilChanged(),
+        filter(val => typeof val === 'boolean')
```
MaxKless, line 39: Why do you filter for differently typed values here? I can't see how other values might get in here. Also, you don't do the same check in the `not-feature-toggle`directive.

src/app/core/directives/not-feature-toggle.directive.ts
```diff
@@ -16,20 +18,43 @@ import { FeatureToggleService } from 'ish-core/utils/feature-toggle/feature-togg
...
   constructor(
     private templateRef: TemplateRef<unknown>,
     private viewContainer: ViewContainerRef,
-    private featureToggle: FeatureToggleService
-  ) {}
+    private featureToggle: FeatureToggleService,
+    private cdRef: ChangeDetectorRef
+  ) {
+    this.enabled$.pipe(distinctUntilChanged(), takeUntil(this.destroy$)).subscribe(enabled => {
+      if (enabled) {
```
MaxKless, line 34: I think it's a little confusing that you check for enabled here when you actually pass `!value` to `enabled$`. 
So either this variable should be called `disabled` or you could pass the normal value to `enabled$` and check for the negation here, in my opinion.
Of course feel free to give me your thoughts on this or correct me if I misunderstood something.

MaxKless (changes requested):
Hi @cesarGamaSa, thanks for this great PR. It looks very good and basically ready to merge, I just have two small questions :)

src/app/core/directives/feature-toggle.directive.ts
dhhyi: `BehaviorSubject` is initialized with `undefined`

src/app/core/directives/not-feature-toggle.directive.ts
cesarGamaSa: Agreed. Applied the first suggested change.

MaxKless: Ok glad that you agree @cesarGamaSa .
Sorry to be so picky, but now the `enabled$` Subject is carrying a value called `disabled`. It doesn't really make sense and still makes it confusing for me. 
It seems more logical that as soon as you negate the result from `featureService.enabled$` the stream and values should be called disabled.

cesarGamaSa: Done :)

cesarGamaSa:
Hey @MaxKless. Is there anything else to develop regarding this refactor?

MaxKless (approved):

PR #960, opened 2021-12-13
refactor: move localization keep patterns where they are needed

dhhyi:
## PR Type

Localization keep patterns are managed in the cleanup script.

## What Is the New Behavior?

Localization keep patterns are managed where they are defined. (with an annotation)

[ ] Yes
[x] No

## Other Information

Hi @SGrueber @suschneider @MaxKless 😃 

I think the `.error.` keep pattern in the `scripts/clean-up-localizations.js` is outdated. I see that all of the error feedback regarding payments comes from the API, there should be no need to keep those translation keys in the PWA (also because we now fetch specific keys directly from the ICM). Regarding anything not payment related? I'm not sure...

Please check if my assumption is correct.

In projects every uselessly translated localization is money 🤑 

[AB#72086](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/72086)

### DIFF
--- a/.github/workflows/dead-code.yml
+++ b/.github/workflows/dead-code.yml
@@ -27,4 +27,4 @@ jobs:
       - name: Find Unused TestBed declarations
         run: |
           npx ts-node scripts/cleanup-testbed
-          git diff --exit-code --raw -p --stat
+          npm run check-no-changes
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -47,16 +47,13 @@ jobs:
           npm i
 
       - name: Code Formatting
-        run: npm run format
-
-      - name: Check No Changes from Formatting
-        run: git diff --exit-code --raw -p --stat
+        run: npx npm-run-all format check-no-changes
 
       - name: .dockerignore Sync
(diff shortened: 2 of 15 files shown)


dhhyi:
- reincluded error pattern as global pattern next to local patterns
- added issue for follow up work: #1032

- [ ] The pattern `/^subject.*/i` introduced by a11651d9170f9cb20fbb6566ba90c8d0dd44b9b1 concerning key `subject.has.no.permission.assigned` needs to be handled

SGrueber:
The pattern /^subject.*/i is needed because in ICM versions lower than 7.10.38 the cost center admin is not allowed to read the list of users for the customer. In this case the translation key 'subject.has.no.permission.assigned' is used to show a corresponding error message on the cost center user assignment page. But the regular expression could be more specific

dhhyi:
> But the regular expression could be more specific

So we can do it like this? 8f71726a99e1e76bf1a465da23da5ed08cdd5daa

shauke (approved):

PR #962, opened 2021-12-14
bugfix: empty checkout receipt page

cesarGamaSa:
## PR Type

[X] Bugfix
[  ] Feature
[  ] Code style update (formatting, local variables)
[  ] Refactoring (no functional changes, no API changes)
[  ] Build-related changes
[  ] CI-related changes
[  ] Documentation content changes
[  ] Application / infrastructure changes
[  ] Other: 

When the receipt page is reached in the checkout (so the order is submitted and the basket is now empty), when the page is refreshed, the receipt is empty, only displaying the progress bar.

## What Is the New Behavior?

The receipt page now has the order submitted when refreshed.

[  ] Yes
[X] No


SGrueber:
Hi César and Danilo, pls let me know if this solution is ok for you now.

cesarGamaSa:
Hi @SGrueber . I'm not quite understanding what this change brings, as the behaviour seems the same as before this commit. Could you please clarify me about this?

SGrueber:
Now there is only 1 effect that takes care of fetching the order after redirect/refresh on checkout receipt page. There are less code changes this way.

SGrueber:
@cesarGamaSa Actually, I wanted to use the effects 'selectOrderAfterRedirect' to get the order. This effect only works if the parameter 'redirect' is set. But now I have noticed that this effect also calls the basket service to save the query params at order - this behavior is unwanted, that's why I have rolled back my changes.

SGrueber (approved):

PR #963, opened 2021-12-15
fix: support the character sequence 'sku' in product SKUs

shauke:
## PR Type

[x] Bugfix

The current `product.route` matcher results in an error page if the products SKU contains the character sequence `sku` e.g. `glaskugel_1`.

Example product detail URL:  `http://localhost:4200/Acer/Glaskugel-skuglaskugel_1-catComputers.897.897_Acer`

Results in a wrong REST API call for the product details: `https://pwa-ish-demo.test.intershop.com/INTERSHOP/rest/WFS/inSPIRED-inTRONICS_Business-Site/rest;loc=en_US;cur=USD/products/gel_1?allImages=true`

skugla**sku**_gel_1_

## What Is the New Behavior?

Routes with product SKUs containing the character sequence `sku` work as expected.

[x] No

### DIFF
--- a/src/app/core/routing/product/product.route.ts
+++ b/src/app/core/routing/product/product.route.ts
@@ -30,7 +30,7 @@ function generateProductSlug(product: ProductView) {
   return slug;
 }
 
-const productRouteFormat = new RegExp('^/(.*)?sku(.*?)(-cat(.*))?$');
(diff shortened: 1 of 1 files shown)


dhhyi:
Fixed and put in a workaround so slugs don't contain the term `-cat`. I think it would be more robust if the category ID was awlays separated at the end like `http://localhost:4200/Acer/Glaskugel-skuglaskugel_1/catComputers.897.897_Acer`

PR #965, opened 2021-12-18
refactor: schematic and lint optimization for default module overrides

dhhyi:
## PR Type

## What Is the New Behavior?

- sort module fields in file overrides
- add components to module all overrides if they exist

[ ] Yes
[x] No


shauke:
I suppose this a useful change for PWA Projects linted with `tslint`?
Our next planed release is PWA 2.0 that will include the switch from `tslint` to `eslint`. Is this a change that should be ported to the `eslint` Branch as well?
Is this fix something you would wish for in a potential PWA 1.4.1 Release even though it is not a bugfix? We are not certain that there will be anything like a PWA 1.5.0.

dhhyi:
> I suppose this a useful change for PWA Projects linted with `tslint`? Our next planed release is PWA 2.0 that will include the switch from `tslint` to `eslint`. Is this a change that should be ported to the `eslint` Branch as well?

The change in the schematics is actually more important for us. But yes, #836 should also include the sorting for module fields in eslint.

> Is this fix something you would wish for in a potential PWA 1.4.1 Release even though it is not a bugfix? We are not certain that there will be anything like a PWA 1.5.0.

Well, it would be a good to clearly communicate a release and maintenance schedule for the Intershop PWA. If it is already there, please link it. I didn't stumble upon it, yet. 😜
For our project we are based on the current 1.4 version and will continue cherry-picking things we need, so we don't *need* a release. Another regular 1.x release as a stepping stone towards 2.0 would be nice, though.

MaxKless (approved):

shauke:
@dhhyi The current release planning is focused only on 2.0. This release will bring `eslint` instead of `tslint` and should also include the changed behavior with a standard rule. But for a short time the `develop` branch will feature the changed behavior also with `tslint` now. 😉

PR #972, opened 2022-01-06
chore: minor project improvements

shauke:
**changes**

* fix VSCode Stylelint integration
* set prettier as default formatter for VSCode and clean up '.vscode/settings.json'
* change post-checkout hook (#875) to only inform about the need to run 'npm install'
* remove disturbing "Check Dead Links For All Files" check from merges to the develop branch
* unintrusive spell check fixes
* CSS class names spell check fixes with code changes
* fix spelling error in .scss files 'quarternary' to 'quaternary'
* display product labels by adding PurgeCSS exception to keep 'product-label' styling

[AB#72496](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/72496)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -40,10 +40,6 @@ jobs:
       - name: Check No Changes
         run: git diff --exit-code --raw -p --stat
 
-      - name: Check Dead Links For All Files
-        if: github.ref == 'refs/heads/develop'
-        run: node docs/check-dead-links
-
       - name: Check Dead Links For Changed Files
         if: github.ref != 'refs/heads/develop'
         run: node docs/check-dead-links origin/develop
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -3,6 +3,19 @@
 
 if git diff-tree -r --name-only --no-commit-id $1 $2 | grep -q package.json
 then
-  git clean -xdf src/**/lazy-* projects/**/lazy-*
-  npm install
+  if [ -z "$INTERSHOP_PWA_AUTO_INSTALL" ]
+  then
+    echo
+    echo --------------------------------------------------------------------------------
+    echo The \'package.json\' contains changes. It could be necessary to run \'npm install\'.
+    echo You can automate this by setting INTERSHOP_PWA_AUTO_INSTALL to "true".
+    echo --------------------------------------------------------------------------------
+    echo
+  fi
+
+  if [ "$INTERSHOP_PWA_AUTO_INSTALL" = "true" ]
+  then
+    git clean -xdf src/**/lazy-* projects/**/lazy-*
+    npm install
+  fi
 fi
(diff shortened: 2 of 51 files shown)


.vscode/settings.json
```diff
@@ -1,14 +1,25 @@
 // Place your settings in this file to overwrite default and user settings.
 {
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.renderWhitespace": "all",
   "editor.formatOnSave": true,
+  "editor.formatOnPaste": false,
   "editor.codeActionsOnSave": {
     "source.fixAll.tslint": true
```
SGrueber, line 8: please add
"editor.codeActionsOnSave": {
		"source.fixAll.stylelint": true
	},

SGrueber (changes requested):

SGrueber (approved):

.github/workflows/docs.yml
```diff
@@ -40,10 +40,6 @@ jobs:
       - name: Check No Changes
         run: git diff --exit-code --raw -p --stat
 
-      - name: Check Dead Links For All Files
-        if: github.ref == 'refs/heads/develop'
-        run: node docs/check-dead-links
```
dhhyi, line 45: bye bye correct project documentation 😅

.husky/post-checkout
```diff
@@ -3,6 +3,11 @@
 
 if git diff-tree -r --name-only --no-commit-id $1 $2 | grep -q package.json
 then
-  git clean -xdf src/**/lazy-* projects/**/lazy-*
-  npm install
+  echo
+  echo --------------------------------------------------------------------------------
+  echo The \'package.json\' contains changes. It could be necessary to run \'npm install\'.
```
dhhyi, line 8: hmm, instead of deactivating it completely, how about reading an environment property to decide if the update should be run. In our project I'm actually pretty fond of this functionality.

src/app/core/facades/checkout.facade.ts
```diff
@@ -155,8 +155,10 @@ export class CheckoutFacade {
     ]).pipe(
       // compare baskets only by shippingMethod
       distinctUntilChanged(
+        // spell-checker: disable
```
dhhyi, line 158: there is a functionality to register specific words for the spell-checker

.github/workflows/docs.yml
shauke: Unfortunately the last dead link failures where mostly false positives that result in a "red" development branch even though there is no problem and especially is the problem not related to the latest merge to the develop branch. They had to be solved by adding an exception to the script.
The task has now become a manual step before releasing, so it no longer blames the latest "innocent" commit. 😉

shauke: So this step should not really change to correct project documentation. 😉

PR #973, opened 2022-01-08
fix: catch and log errors from translation API

### LINKED ISSUE #971: localization keys are rendered instead of localization values in case of 'localizations' REST API errors
## Actual Behavior

If the `localizations` REST call fails for some reason when initializing the PWA the localization keys are rendered in the storefront instead of the localization values that should still be available from `\src\assets\i18n\en_US.json`.

image

## Expected Behavior

The information from the `localizations` REST call should only update the default/fallback values from the localization files. If anything goes wrong the default/fallback values need to be rendered in the storefront but not the localization keys.

## Additional Note

Also if the ICM server is not reachable at all the localization keys are rendered instead of the default/fallback localization values.

image

[AB#72443](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/72443)

dhhyi:
## PR Type

[x] Bugfix

Issue Number: Closes #971

## What Is the New Behavior?

- error caught, logged and empty translations returned
- testing

[ ] Yes
[x] No

## Other Information

[AB#72527](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/72527)

### DIFF
--- a/src/app/core/services/localizations/localizations.service.spec.ts
+++ b/src/app/core/services/localizations/localizations.service.spec.ts
@@ -1,21 +1,81 @@
-import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
+import { ErrorHandler } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { provideMockStore } from '@ngrx/store/testing';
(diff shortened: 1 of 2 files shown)


MaxKless (approved):
Nice and thanks for the tests

PR #1007, opened 2022-02-09
feat: product context with propagateActive as input

cesarGamaSa:
## PR Type

[ ] Bugfix
[X] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Propagation is done when changes are applied which include the propagateActive flag.

## What Is the New Behavior?

propagateActive can also be changed as an input of the product context directive, which affects the propagation timing and availability.

[ ] Yes
[X] No


MaxKless:
Hi @cesarGamaSa, thanks for the PR. 
I see that this change is just one line in the product context directive. I assume you are using this somewhere in your project and need this change to make your life easier there?
Is there somewhere you can use this new change in a template in the standard PWA? Because as of now, this would be kind of like merging "dead code" that we don't use ourselves.

I'm not saying I won't merge this if you can't find a location I'm just curious :)

cesarGamaSa:
Hello @MaxKless. In our project we added this because we found a huge utility in it. We've made a custom quick order which uses product context and two-way data binding for SKU and quantity. In order to enable the add to cart button only when we have at least an SKU with some quantity, we added this and it worked perfectly and made this implementation a lot easier. So we thought that this could also be useful for the standard PWA, ence my PR.

MaxKless (approved):

PR #1018, opened 2022-02-17
feat: support personalized product prices

Eisie96:
## PR Type

[x] Feature

Product prices are fetched with the products REST resource.

## What Is the New Behavior?

The PWA uses the new productprice REST API to get the (personalized) product prices.
On Logout or Login all personalized prices are removed from store.

[x] Yes

BREAKING CHANGES: Product prices are now fetched by separate `productprices` REST calls instead of using the information fetched with the products REST call.

## Other Information

[AB#74406](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/74406)

### DIFF
--- a/src/app/core/services/prices/prices.service.spec.ts
+++ b/src/app/core/services/prices/prices.service.spec.ts
@@ -1,0 +1,54 @@
+import { TestBed } from '@angular/core/testing';
+import { provideMockStore } from '@ngrx/store/testing';
+import { of } from 'rxjs';
+import { anything, capture, instance, mock, verify, when } from 'ts-mockito';
+
+import { Customer } from 'ish-core/models/customer/customer.model';
+import { ProductPriceDetailsData } from 'ish-core/models/product-prices/product-prices.interface';
+import { ApiService, AvailableOptions } from 'ish-core/services/api/api.service';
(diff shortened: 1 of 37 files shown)


src/app/core/services/prices/prices.service.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      when(apiServiceMock.get<{ data: ProductPriceDetailsData[] }>('productprices', anything())).thenReturn(
+        of({ data: [{ sku: 'abc' }, { sku: '123' }] as ProductPriceDetailsData[] })
+      );
+    });
+
+    it('should get product price data when "getProductPrices" is called', done => {
+      pricesService.getProductPrices(['abc', '123']).subscribe(priceDetails => {
+        expect(priceDetails).toHaveLength(2);
+        verify(apiServiceMock.get(`productprices`, anything())).once();
+        expect(capture(apiServiceMock.get).last()).toMatchInlineSnapshot(`
```
MaxKless, line 47: This looks a little weird to me. Why do we care about all the specific properties from the `ApiService` here? 
In this place I would test whether the `PricesService` returns the correct details, don't you agree?

MaxKless (changes requested):

PR #1030, opened 2022-02-28
feat: handle and display scaled prices

suschneider:
## PR Type
[x] Feature

Scaled prices are retrieved from Product Pricing REST API, but are not displayed in the storefront.

## What Is the New Behavior?
Scaled prices are displayed on the product detail page, product listings, etc.

[ ] Yes
[x] No

[AB#74617](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/74617)


Eisie96 (commented):
showScaledPrice input value is not set for product-compare-list.component. I think it is a good idea to show scaled prices for a compare between products to help the customer.

M-Behr (changes requested):
Nice feature! I found a little issue with the quote request.

1. Go to product detail page
2. Set quantity to 3
3. Click "add to quote request"
4. Change the quantity to 1 on your quote request
--> The price for 3 items is still displayed

Expected: It should display the updated price depending on the selected quantity (like on the cart page) - there the scaled price info is not needed (that's fine).

Note: The other behavior is to display the price for one item and the scaled price info (see pdp or order templates).

iwiederhold (dismissed):
👍

Eisie96:
Thank you, @M-Behr, for reporting the bug. The reason for this behaviour is, that the quote request API doesn't update correctly the prices for the quote line items. Therefore the bug has nothing to do with the new scaled price feature. Nevertheless the bug was created and will be solved soon.

M-Behr (approved):
Good work 👍

shauke (approved):

PR #1055, opened 2022-03-09
feat(eslint): rules for customization

dhhyi:
## PR Type

[x] Feature
[x] Other: Best Practice

When using the Intershop PWA with multiple themes, some best practices should be honored:
- the `THEME` variable should not be used in components, instead component file overrides should be used
- properties `templateUrl` and `styleUrls` should always point to the base file, so the override mechanism can step in at build time.

## What Is the New Behavior?

Lint rules added.

[ ] Yes
[x] No

## Other Information

[AB#75105](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75105)

### DIFF
--- a/eslint-rules/src/rules/do-not-use-theme-identifier.ts
+++ b/eslint-rules/src/rules/do-not-use-theme-identifier.ts
@@ -1,0 +1,41 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+
+import { normalizePath } from '../helpers';
+
+const DEFAULT_FILE_PATTERN = '.*\\.(component|directive|pipe)\\.ts';
+
+export const doNotUseThemeIdentifierRule: TSESLint.RuleModule<string, [string]> = {
+  meta: {
+    docs: {
+      description: `Using the THEME variable in Angular artifacts directly bypasses the concept of component overrides and leads to bad practice. This rule warns about the use. The pattern for files can be configured. Default is "${DEFAULT_FILE_PATTERN}"`,
+      recommended: 'warn',
+      url: '',
+    },
+    messages: {
+      doNotUseThemeIdentifier: 'Do not use THEME variable here. Instead use correctly named overrides.',
+    },
+    type: 'problem',
+    schema: [
+      {
+        type: 'string',
+        additionalProperties: false,
+      },
+    ],
+  },
+  create(context) {
+    const config = context.options[0] ?? DEFAULT_FILE_PATTERN;
+
+    if (!new RegExp(config).test(normalizePath(context.getFilename()))) {
+      return {};
+    }
+
+    return {
+      'Identifier[name="THEME"]'(node) {
+        context.report({
+          node,
+          messageId: 'doNotUseThemeIdentifier',
+        });
+      },
+    };
+  },
+};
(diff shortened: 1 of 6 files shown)


eslint-rules/src/rules/use-correct-component-overrides.ts
```diff
@@ -0,0 +1,136 @@
...
+      pointToBasicFile: 'Replace with basic component file.',
+    },
+    type: 'problem',
+    schema: [],
+    hasSuggestions: true,
+  },
+  create(context) {
+    const fileName = path.basename(normalizePath(context.getFilename()));
+    const fileBaseName = fileName.replace(/component.*/, 'component');
+    if (/\.component\.(?!(spec|stories))([a-z]+\.)?ts$/.test(fileName)) {
```
MaxKless, line 24: Maybe you can make this regexp configurable via `eslintrc.json`? It doesn't really make sense to include `stories` here for the PWA, even though some projects might use it ;)

eslint-rules/src/rules/use-correct-component-overrides.ts
```diff
@@ -0,0 +1,136 @@
...
+            context.report({
+              node,
+              messageId: 'shouldPointToBasicFile',
+              suggest,
+            });
+          }
+        },
+      };
+    } else if (/\.component\.[a-z]+\.spec\.ts$/.test(fileName)) {
+      function classify(base: string): string {
```
MaxKless, line 61: I feel like we have too many implementations of these similar string manipulation functions in the PWA. I think it would be nice to at least put this method in the `helpers.ts` so people coming after won't add yet another implementation. 
Btw, why not use `strings.classify` from `@angular-devkit/core` like we do for schematics?

eslint-rules/src/rules/use-correct-component-overrides.ts
```diff
@@ -0,0 +1,136 @@
+import { Selectors } from '@angular-eslint/utils';
+import { ASTUtils, TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { normalizePath } from '../helpers';
+
+export const useCorrectComponentOverridesRule: TSESLint.RuleModule<string, []> = {
```
MaxKless, line 8: Please add some documentation to explain the rule, what it does and how to configure it.

eslint-rules/src/rules/do-not-use-theme-identifier.ts
```diff
@@ -0,0 +1,34 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+
+import { normalizePath } from '../helpers';
+
+export const doNotUseThemeIdentifierRule: TSESLint.RuleModule<string, [string]> = {
```
MaxKless, line 5: Please add some documentation to explain the rule, what it does and how to configure it.

MaxKless (changes requested):
Looks mostly good :)

eslint-rules/src/rules/use-correct-component-overrides.ts
dhhyi: Just realized that neither spec nor stories will define a Component with URL fields. But `spec` is needed for the if/else.

eslint-rules/src/rules/use-correct-component-overrides.ts
dhhyi: Good idea using `strings` 👍

MaxKless (approved):

PR #1056, opened 2022-03-09
feat: desired delivery date

### LINKED ISSUE #698: Desired delivery date missing

On PWA, on shipping step of funnel,  the field with the calendar that allows customer to fill his required shipping date are missing

Is it on your roadmap to have on PWA version the same feature as on classic version ?

calendar_delivery_step

This feature is needed on our project

shauke:
## PR Type

[x] Feature

There is no possibility to select a desired delivery date during checkout. 

Issue Number: Closes #698

## What Is the New Behavior?

A desired delivery date can be selected for the whole basket at the shipping page during checkout.

[x] No

## Other Information

The following tasks cannot be done before the REST api supports the setting of the desired delivery date at the basket line items. See  [AB#75181](https://dev.azure.com/intershop-com/Products/_workitems/edit/75181)

- [ ] save ddd at basket line items immediately after the date has been entered
- [ ] check whether all basket items have the ddd set correctly before submitting the basket (before the final basket validation) and update line items if necessary

[AB#75122](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75122)

- [ ] add ddd to the e-mail templates
- [x] reduce the input field length by half ("col-md-5") - also for the order reference ID -> seems that the shipping page need a restructuring for that


src/assets/i18n/en_US.json
```diff
@@ -650,6 +650,14 @@
   "checkout.credit_card.user.lastname.missing.error": "Please enter a cardholder last name.",
   "checkout.credit_card.user.name.missing.error": "Please enter a cardholder name.",
   "checkout.credit_card.validityTime.error.notFound": "The maximum validity time for CVC has not been found.",
+  "checkout.desired_delivery_date.apply.button.label": "Apply",
+  "checkout.desired_delivery_date.error.date": "Invalid date. Please use this format (yyyy-mm-dd).",
```
M-Behr, line 654: The date is invalid. Please use the correct format (yyyy-mm-dd).

M-Behr (changes requested):
The desired delivery date works fine. Here are some points, which I found during my review:

- [ ] I would like to change the error messages a little bit. See comments in the file.
- [ ] Please add the Germany and French translations. Don't forget to localize the date format. Ask @mglatter for the localization :)
- [ ] Decrease the space between error message and input field in phone view (see screenshot).
Spaces_error_message_input field

- [ ] Add help text "YYYY-MM-DD" below input field, so that the user knows how he has to enter a date. See help text style in Order Reference ID field (see screenshot). Note: I thought we can use the placeholder for this, but a help text is better. I changed this in the IAD (Figma file) as well.
image

- [ ] Use the primary-color for the selected date in the date picker.
- [ ] Use text color  #6c757d for the weekdays in the date picker. It's the color of the bootstrap default class "text-muted".
- [ ] Please check if the desired delivery date is also displayed in our e-mail templates (maybe this is ICM related and out of scope).
- [ ] Reduce the input field length by half. "col-md-5" should be fine for desktop. Maybe @SGrueber could do this because this adaption is also needed for the order reference ID. This field is also too large in desktop view.

MaxKless (changes requested):
Looks pretty good already, just a bunch of small stuff.

johnopaluwa:
> 

@mglatter should I send the translation texts on teams or you will like to make the changes directly on git?

src/app/core/utils/date-helper.ts
```diff
@@ -0,0 +1,86 @@
+import { NgbDateStruct } from '@ng-bootstrap/ng-bootstrap';
+import { Observable, isObservable, map, of } from 'rxjs';
+
+export class DateHelper {
+  private static readonly DELIMITER = '-';
+
+  static addDays(date: Date, days: number) {
+    const result = new Date(date);
+    result.setDate(date.getDate() + days);
```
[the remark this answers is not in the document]
johnopaluwa: I tried to imagine it in one line and what I came with was 
```new Date(date).setDate(date.getDate() + days)``` .
The above expression returns a number and not a date.

If I also try to do
 ```date.setDate(date.getDate() + days)```
 I will be mutating the argument pass into the function which is a behaviour I dont wish to have.

File: src/assets/i18n/en_US.json [hunk 3]
```diff
@@ -650,6 +650,16 @@
   "checkout.credit_card.user.lastname.missing.error": "Please enter a cardholder last name.",
   "checkout.credit_card.user.name.missing.error": "Please enter a cardholder name.",
   "checkout.credit_card.validityTime.error.notFound": "The maximum validity time for CVC has not been found.",
+  "checkout.desired_delivery_date.apply.button.label": "Apply",
+  "checkout.desired_delivery_date.error.date": "The date is invalid. Please use the correct format (yyyy-mm-dd).",
+  "checkout.desired_delivery_date.error.max_date": "Your desired delivery date is excluded because the date is too far in the future.",
+  "checkout.desired_delivery_date.error.min_date": "Your desired delivery date is excluded because the date is in the past or falls below the minimum delivery time.",
+  "checkout.desired_delivery_date.error.no_saturday": "Your desired delivery date is excluded because deliveries are not made on Saturdays.",
+  "checkout.desired_delivery_date.error.no_sunday": "Your desired delivery date is excluded because deliveries are not made on Sundays.",
+  "checkout.desired_delivery_date.label": "Desired delivery date",
```
M-Behr, line 659: Labels are always displayed in title case - Desired Delivery Date.

src/app/shared/components/basket/basket-shipping/basket-shipping.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BasketShippingComponent {
+  @Input() set object(val: Basket | Order | Requisition) {
+    this.object$.next(val);
+  }
+
+  readonly object$ = new ReplaySubject<Basket | Order | Requisition | undefined>(1);
+
+  readonly desiredDeliveryDate$ = this.object$.pipe(
```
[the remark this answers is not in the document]
johnopaluwa: Thanks

src/app/shared/formly/types/date-picker-field/date-picker-field.component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  get minDate$(): Observable<NgbDateStruct> {
+    return DateHelper.addDaysToTodayAsync(this.to.minDays);
+  }
+
+  get maxDate$(): Observable<NgbDateStruct> {
+    return DateHelper.addDaysToTodayAsync(this.to.maxDays);
+  }
+
+  markDateDisable(isSatExcluded: unknown, isSunExcluded: unknown) {
```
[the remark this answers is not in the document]
johnopaluwa: because then tslint would complain about this ```const disableSaturday = isSatExcluded === true ? DateHelper.isSaturday(date) : false;``` => unnecessarily compares a boolean

however this statement  ```isSatExcluded === true```  is a guard that needs to be there  to sanitize the input.

and having an expression like ```!!isSatExcluded ``` would fail as a guard because ```!!(valid number)``` will also be true.

src/app/shared/formly/types/date-picker-field/date-picker-field.component.html
```diff
@@ -0,0 +1,28 @@
...
+    [maxDate]="maxDate$ | async"
+    [markDisabled]="
+      markDateDisable(
+        isObservableFn(this.to.isSatExcluded) ? (this.to.isSatExcluded | async) : this.to.isSatExcluded,
+        isObservableFn(this.to.isSunExcluded) ? (this.to.isSunExcluded | async) : this.to.isSunExcluded
+      )
+    "
+    outsideDays="collapsed"
+    navigation="arrows"
+    placeholder="mm/dd/yyyy"
```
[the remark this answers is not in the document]
johnopaluwa: I will take out the place holder and replace with a help text. That was a feedback from @M-Behr .

mglatter:
> > 
> 
> @mglatter should I send the translation texts on teams or you will like to make the changes directly on git?

@johnopaluwa I can look for the respective json files in Github and directly add the translations on Git. I would only need a signal from you once the texts are final.

src/app/shared/formly/types/date-picker-field/date-picker-field.component.scss
```diff
@@ -0,0 +1,6 @@
+.input-group {
+  // hack to change the days of week color
+  ::ng-deep .ngb-dp-weekday {
+    color: #6c757d !important;
```
shauke, line 4: I would not call it a hack to use `ng-deep`. We explicitly allowed it in the the Stylelint configuration (https://github.com/intershop/intershop-pwa/blob/develop/.stylelintrc.json#L25). This is fine this way.
What is actually a hack is the hard coded color.
You can use `@import 'variables';` in the beginning of the `.scss` file and with that you have access to all our variables (but not bootstrap variables directly). `$text-muted` maps to `$gray-600` so you can write `color: $gray-600 !important;`.
For such a gray value this might no be as important but if you want to work with the primary color for example you cannot hard code it here since it will be used in different themes that might have different primary colors.

src/assets/i18n/de_DE.json
```diff
@@ -650,6 +650,16 @@
   "checkout.credit_card.user.lastname.missing.error": "Bitte geben Sie den Nachnamen des Karteninhabers an.",
   "checkout.credit_card.user.name.missing.error": "Bitte geben Sie den Namen des Karteninhabers an.",
   "checkout.credit_card.validityTime.error.notFound": "Die maximale Gültigkeitsdauer der Kartenprüfnummer wurde nicht gefunden.",
+  "checkout.desired_delivery_date.apply.button.label": "Übernehmen",
+  "checkout.desired_delivery_date.error.date": "Das Datum ist ungültig. Bitte verwenden Sie das korrekte Format (TT.MM.JJJJ).",
```
mglatter, line 654: Please check if the date picker plugin has localized date formats. Else the given format in our messages has to be changed accordingly.

src/assets/i18n/fr_FR.json
```diff
@@ -650,6 +650,16 @@
   "checkout.credit_card.user.lastname.missing.error": "Veuillez entrer le nom de famille du titulaire de la carte.",
   "checkout.credit_card.user.name.missing.error": "Veuillez entrer le nom du titulaire de la carte.",
   "checkout.credit_card.validityTime.error.notFound": "La durée de validité maximale du CVC n’a pas été trouvée.",
+  "checkout.desired_delivery_date.apply.button.label": "Appliquer",
+  "checkout.desired_delivery_date.error.date": "La date n’est pas valide. Veuillez utiliser le format correct (jj-mm-aaaa).",
```
mglatter, line 654: Please check if the date picker plugin has localized date formats. Else the given format in our messages (lines 654 and 660) has to be changed accordingly.

src/assets/i18n/de_DE.json
```diff
@@ -781,6 +792,7 @@
   "customer.credentials.passwordreset.invalid_password.error.PasswordExpressionViolation": "Das Kennwort muss mindestens 7 Zeichen enthalten, bestehend aus Zahlen und Buchstaben, ohne Leerzeichen.",
   "customer.missing_fields.error": "Fehlende Felder",
   "customer.wishlist.profane_word.error": "Ihre aktuelle Eingabe kann nicht bearbeitet werden. Unpassende, beleidigende bzw. vulgäre Sprache wird nicht akzeptiert. Bitte geben Sie etwas anderes ein.",
+  "date_format": "dd.MM.yyyy",
```
dhhyi, line 795: That should not be part of the translation file, as it is part of the Angular localization (sadly, internal API). Get it with:

```ts
import { ɵfindLocaleData } from '@angular/core';

const localeData = ɵfindLocaleData(locale);
return localeData[10][0].replace(/\by\b/, 'yyyy');

```

src/assets/i18n/fr_FR.json
```diff
@@ -650,6 +650,16 @@
...
+  "checkout.desired_delivery_date.error.max_date": "La date de livraison souhaitée ne peut être choisie car elle est trop éloignée dans le temps.",
+  "checkout.desired_delivery_date.error.min_date": "La date de livraison souhaitée ne peut pas être choisie car elle se situe dans le passé ou est inférieure à la date de livraison la plus proche possible.",
+  "checkout.desired_delivery_date.error.no_saturday": "La date de livraison souhaitée ne peut être choisie car les livraisons ne sont pas effectuées le samedi.",
+  "checkout.desired_delivery_date.error.no_sunday": "Votre date de livraison souhaitée ne peut pas être choisie car les livraisons ne sont pas effectuées le dimanche.",
+  "checkout.desired_delivery_date.label": "Date de livraison souhaitée",
+  "checkout.desired_delivery_date.note": "JJ-MM-AAAA",
```
dhhyi, line 660: I don't think date format placeholders should be part of the translation file, as they can be derived from the standard shortDate format, which can be retrieved from there.

Maybe (in this case) add 3 general translation keys for replacing the letters:
```
 "common.date_format.letters.y": "A",
 "common.date_format.letters.M": "M",
 "common.date_format.letters.d": "J", 
``` 
And then use the shortDate format to transform it. (See other comment)

src/app/shared/formly/types/date-picker-field/date-picker-field.component.scss
```diff
@@ -0,0 +1,8 @@
+@import 'variables';
+
+.input-group {
+  ::ng-deep .ngb-dp-weekday {
```
dhhyi, line 4: `::ng-deep` is deprecated, isn't it? Global styling should be used here.

dhhyi (changes requested):

src/app/shared/formly/types/date-picker-field/date-picker-field.component.scss
shauke: Looking at our CoP Angular discussion results we agreed on using `::ng-deep` in this way in component styles:
```
 :host ::ng-deep
```
That way it gets some kind of encapsulation back.

But @dhhyi is right this could also be a global style, especially since we probably want all our date pickers to look the same not just the once used via formly. What about adding `.ngb-dp-weekday` to `src\styles\global\forms.scss`?

PR #1060, opened 2022-03-10
build: deploy-url deprecated during docker build

dhhyi:
## PR Type

[x] Build-related changes

Parameter `--deploy-url` is deprecated

[ ] Yes
[ ] No


shauke:
We probably should remove the deploy-url functionality since it is deprecated and not really needed by the projects?
The approach of this PR is not the right way though.

PR #1063, opened 2022-03-10
refactor: product context improvements

dhhyi:
## PR Type

[x] Bugfix
[x] Feature
[x] Refactoring (no functional changes, no API changes)

## What Is the New Behavior?

Minor improvements to the product context.

## Does this PR Introduce a Breaking Change?

[ ] Yes
[x] No

## Other Information

[AB#75164](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75164)

Issue #748, opened 2021-06-28
Retail sets are not split into line items when added to lists


M-Behr:
**Actual Behavior**
When a retail set is added to a list (e.g. quotes, order templates, wish list), the retail set is displayed in the list without a quantity. Additionally it is not possible to add a retail set to cart as quote item. Error code: 400.

**Expected Behavior**
When a retail set is added to a list (e.g. quotes, order templates, wish list), it must be divided into line items. The line items have a quantity field and it is possible to add them to the cart. See Responsive Starter Store as reference. 

**Steps to Reproduce the Bug**

1. Go to 'Homepage'
2. Scroll down to 'Microsoft Surface Pro Set'
2. Click on 'Add to Quote Request' icon or 'Add to Order Template' icon
3. See product in the list

**Environment Details**

  - Browser: Chrome
  - Public Demo Server

**Additional Context, like Screenshots, Log File Snippets etc.**
Current:
![2021-06-28 10_15_58-My Account _ Intershop PWA](https://user-images.githubusercontent.com/66066967/123604674-5b181480-d7fb-11eb-965d-2c41da8afb6d.png)
Expected:
![2021-06-28 10_15_27-My Account - Quote Details](https://user-images.githubusercontent.com/66066967/123604698-610df580-d7fb-11eb-94e5-cab0d65c48fa.png)

SGrueber:
This functionality has to be provided by the ICM server.

shauke:
The existing quantity handling issues when adding retail sets to quotes or order templates where handled in the PWA (#2146, #2160).
The actual splitting of retail sets when adding them to the lists will be handled in ICM.
Therefore this ticket can be closed now.

PR #1066, opened 2022-03-11
fix: disable add-to-basket when quantity is zero

dhhyi:
## PR Type

[x] Bugfix

"Add to Cart" buttons on retail sets are active when quantity is 0.

Steps to reproduce:
- Navigate to a retail set (i.e. https://intershoppwa.azurewebsites.net/PCs/Fujitsu-ESPRIMO-Edition-skuM8540840-catComputers.1835.153)
- set an individual quantity of a retail set part to 0
  Expected: Add to cart should not be possible
  Actual: Add to cart button is enabled, clicking it will lead to an endless loading animation
- set all quantities to 0
  Expected: All add to cart buttons should be disabled
  Actual: Add to cart button is enabled, clicking it will lead to an endless loading animation

## What Is the New Behavior?

- disable add to cart for quantity below 1
- track quantity of children in parent context to enable the same behavior

[ ] Yes
[x] No

## Other Information

WAIT: Fix is based on #1063, please review and merge before this fix.

[AB#76057](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76057)

### DIFF
--- a/src/app/core/facades/product-context.facade.ts
+++ b/src/app/core/facades/product-context.facade.ts
@@ -260,6 +260,25 @@ export class ProductContextFacade extends RxState<ProductContext> implements OnD
     );
 
     this.connect(
+      'quantity',
+      this.select('children').pipe(
+        map(children => Object.values(children)),
+        skipWhile(children => !children?.length),
+        map(children =>
+          children.reduce(
+            (sum, child) =>
+              sum +
+              (Number.isInteger(child.quantity) && !child.hasQuantityError && !child.hasProductError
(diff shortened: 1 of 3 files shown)


dhhyi:
@Eisie96 sorry, I misunderstood your comment. Yes, if a retail set is added to an order template from the detail page, the quantities are wrong. I'll fix something. 😅

dhhyi:
@Eisie96 now based on #1141 which handles #748

PR #1080, opened 2022-03-22
refactor: introduce SSR variable for platform checks

dhhyi:
## PR Type

[x] Feature

For checking if the current Angular app is running in SSR or Browser, the `PLATFORM_ID` has to be injected and the methods `isPlatformBrowser` and `isPlatformServer` have to be used to perform the check. This always requires DI context and is rather verbose.

## What Is the New Behavior?

- The customized build sets the variable `SSR` that can be used for this check.
- Added lint rule that suggests code changes.
- Transformed all occurrences.
- Add `onSSREnvironment` to jest `describe` so the SSR variable can be set for testing (default is false)

[ ] Yes
[x] No

## Other Information

As the `SSR` value is now injected into the build process with the [webpack DefinePlugin](https://webpack.js.org/plugins/define-plugin/), the values can be used for code minification. When using the usual way with `PLATFORM_ID`, the chack can only be performed at runtime and all code is ending up in server and client bundles.

Verify this by adding the following code:
```ts
    if (!SSR) {
      console.log('browser - QWERTY1234');
    }
    if (SSR) {
      console.log('server - QWERTY1234');
    }
```

Check where it turns up with (bash):
```
~$ npm run build
...
~$ grep -ro '\w* - QWERTY1234' dist
dist/server/main.js:server - QWERTY1234
dist/browser/main.e9860fd746f0e9c6.js:browser - QWERTY1234
```

[AB#75426](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75426)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -657,6 +657,7 @@
         "ish-custom-rules/use-component-change-detection": "warn",
         "ish-custom-rules/use-correct-component-overrides": "warn",
         "ish-custom-rules/use-jest-extended-matchers-in-tests": "warn",
+        "ish-custom-rules/use-ssr-variable-instead-of-platform-id": "warn",
         "ish-custom-rules/require-formly-code-documentation": "warn",
         "jest/no-commented-out-tests": "warn",
         "jest/no-disabled-tests": "warn",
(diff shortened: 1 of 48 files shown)


shauke:
It seems like a nice improvement but we should introduce it with the next major version and not with a 2.x minor version. At least if there are no further objections.

dhhyi:
> It seems like a nice improvement but we should introduce it with the next major version and not with a 2.x minor version.

Good idea. I'll keep the branch up-to-date until then whenever it `has conflicts` 😉👍

shauke (approved):

PR #1084, opened 2022-03-24
refactor: compile schematics into a dist folder

dhhyi:
## PR Type

[x] Build-related changes

Generated sources in schematics cluttered into source folder.

## What Is the New Behavior?

A dedicated dist folder is used.

[ ] Yes
[x] No


dhhyi:
Please merge #1078 first

schematics/src/address-form-configuration/factory_spec.ts
```diff
@@ -1,10 +1,9 @@
 import { UnitTestTree } from '@angular-devkit/schematics/testing';
 import { lastValueFrom } from 'rxjs';
 
+import { PWAAddressFormConfigurationOptionsSchema as Options } from '../../dist/address-form-configuration/schema';
```
MaxKless, line 4: Why are the inputs inconsistent here? Sometimes you use `../../dist` and sometimes `schemas/`. Shouldn't jest be able to use the same paths because of the `moduleNameMapper`?

dhhyi: Yes, it will work but then VSCode cannot resolve the import because the TS path mapping is not in the root `tsconfig.json` but in the one in the schematics folder that is not picked up.

So two options:
1. consistency but false-negative VSCode compile errors and no IDE code support
2. relative imports from the dist folder in tests (which is also a bit confusing, but works)

MaxKless (approved):

PR #1087, opened 2022-03-28
chore: prefix style bundle with theme

dhhyi:
## PR Type

[x] Build-related changes

## What Is the New Behavior?

Style bundles are prefixed with the theme:
image

This makes it easier to distinguish the build table when building in parallel.

[ ] Yes
[x] No

## Other Information

[AB#75548](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75548)


shauke:
Nice to see what is all possible but we would pass to add it as an standard feature.
Thank your for showing us what can be done with the custom webpack builder. 😉

PR #1089, opened 2022-03-29
feat: introduce price update strategies

dhhyi:
## PR Type

[x] Feature

Prices are fetched whenever the `LoadProductSuccess` action is encountered. This happens also if the actual price is not being displayed.

## What Is the New Behavior?

- only fetch prices that are displayed (via context displayProperties)
- introduced two update strategies:
  - *stable*: fetch prices only once per application lifetime
  - *always*: re-fetch prices every time they are displayed
- when going to Product detail pages, the prices are updated if the product data is also updated (when completeness level `true` is chosen in `SelectedProductContextFacade`)

[ ] Yes
[x] No


dhhyi:
wait for integration of #1088

Eisie96 (approved):

PR #1096, opened 2022-03-30
perf: faster npm install

dhhyi:
## PR Type

[x] Build-related changes

## What Is the New Behavior?

`npm install` is executed with `--prefer-offline` and `--no-audit` for `clean` scripts and docker builds.

[ ] Yes
[x] No

## Other Information

[AB#75600](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75600)


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:16-alpine as buildstep
 ENV CI=true
 WORKDIR /workspace
 COPY package.json package-lock.json /workspace/
-RUN npm i --ignore-scripts
+RUN npm ci --prefer-offline --no-audit --ignore-scripts
```
shauke, line 6: How can any `npm ci` be faster then `npm i`? Only if there is nothing to clean anyways, right?
Is the switch to `ci` in the Docker files intentionally?

shauke (changes requested):

Dockerfile
dhhyi: Yes, docker builds should always use `npm ci`, because then it will fail if the lock-file is not in sync with the package.json.
The `--no-audit` switch makes it a bit faster. And you are right: docker builds never have anything to clean.

shauke (approved):

PR #1105, opened 2022-04-04
fix: hybrid approach adaptions

shauke:
## PR Type

[x] Bugfix

Hybrid approach did not work for anonymous users (missing functionality on ICM side).

## What Is the New Behavior?

* adapted `apiToken` cookie handling after ICM adaptions (ICM version ???) for anonymous baskets
* fix static file handling for CSS/JS files from ICM
* adapt mapping table

[x] No

## Other Information

Needs ICM 7.10.38.6-LTS or 7.10.32.16-LTS or newer.

[AB#75730](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/75730)


nginx/templates/multi-channel.conf.tmpl
```diff
@@ -126,6 +126,26 @@ server {
...
+        proxy_set_header Host $http_host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+
+        proxy_no_cache true;
+        proxy_cache_bypass true;
+        add_header X-Cache-Status $upstream_cache_status;
+
+        proxy_pass {{ getenv "ICM_BASE_URL" }};
```
dhhyi, line 143: With that the `ICM_BASE_URL` becomes mandatory as an environment variable for nginx, which it shouldn't.

shauke: It now has a condition around this block. Does this look better?

By the way why does the next block use `$UPSTREAM_PWA` while it seams to do pretty much the same?
```
    # let ICM handle everything ICM related
    location ~* ^/INTERSHOP.*$ {
    ...
        proxy_pass {{ $UPSTREAM_PWA }};
    ...
```

dhhyi: Yes, with the condition it works.

The next block is used for the `PROXY_ICM` feature in the PWA SSR container, that also activates when `SSR_HYBRID` is active. So the question is, why do you actually need the additional block in the setup to go directly to the ICM instead of just letting the PWA SSR delegate it.

What was the original error? It couldn't resolve static files? CORS errors?

SGrueber:
The redirect after checkout still works as expected.

nginx/templates/multi-channel.conf.tmpl
shauke: statc_INTERSHOP_resources

I would say without this additional block it couldn't resolve static files and not CORS issues.
And it is probably right to directly let ICM handle the static files and leave out the SSR container also for performance reasons.

dhhyi (dismissed):

nginx/templates/multi-channel.conf.tmpl
dhhyi: 👍

PR #1106, opened 2022-04-04
fix: checkout payment submit disabled check

carlosDjangoo:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The checkout payment submit disabled check is wrong. The button is not disabled when something is wrong in the form.

Issue Number: Closes #

## What Is the New Behavior?

The checkout payment submit disabled check is correct.

[ ] Yes
[x] No

### DIFF
--- a/src/app/pages/checkout-payment/checkout-payment/checkout-payment.component.ts
+++ b/src/app/pages/checkout-payment/checkout-payment/checkout-payment.component.ts
@@ -250,7 +250,7 @@ export class CheckoutPaymentComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   get submitDisabled() {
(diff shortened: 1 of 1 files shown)


SGrueber:
Our form behavior works the way, that the submit button is enabled even if there are validation errors in the form. Only after the user has clicked the submit button all validation errors of the form are shown and the submit button is disabled.
All of our forms have a similar code snippet: 
```
get formDisabled() {
    return this.form.invalid && this.submitted;
 }
```
You will find more information about our form handling here (section 'Form Handling'): 
https://github.com/intershop/intershop-pwa/blob/develop/docs/guides/forms.md

I am sorry, but we cannot accept your changes.

PR #1108, opened 2022-04-07
feat: map errors into an object

ivopereira15:
## PR Type

[ X] Bugfix
[ X] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

At this moment when we have more than one error we are showing an warning on browser console. This is a problem in the quick order because when we have more than one product with error/errors we dont show feedback to the customer.
Also when we are showing the error we dont know which product we are having the error.

## What Is the New Behavior?
With this change we are fixing two problems
- Show error when we have more than one
- Have the possibility to know which product is having an error.

We are adding a property named `causes` to the HttpError. This property will have the error message with more details and if it has will have the product sku.

Examples 
Quick order when we are trying to add two products
quickorder_tworproducts

[ ] Yes
[ X] No


dhhyi (dismissed):

dhhyi (dismissed):

M-Behr:
Thank you for your feedback. I see 3 variants to improve the error case behavior:

Var 1:
2 products could not be added. (--> general error)
The product quantity for these items in your cart... (--> reason summary for same errors)

Var 2:
2 products could not be added. (--> general error)
The product quantity for "Microsoft Surface Hub" in your shopping cart... (--> reason 1)
The product quantity for "Microsoft Surface Book" in your shopping cart... (--> reason 2)

Alternative: The product quantity for product with SKU "123456" in your shopping cart...
(Is this really helpful because we don't show SKUs in mini cart?)

Var 3:
If error handling is hard to overview in mini cart, we could display always shopping cart page in cases of errors or adjustments (see Wireframes).
image

We will check your solution and discuss if it's fits to our interaction guidelines.

M-Behr:
I discussed the error behavior with @SGrueber.
We want to jump to the cart page if we get multiple product errors after adding products to the cart. There we can display the individual errors directly on the product and make it much clearer. It should be look like that:

2022-04-25 13_41_10-Failed Inventory Check

We would continue to display error messages for a single product in the mini shopping cart. We would leave the behavior for this case as it is.

@ivopereira15 @dhhyi Is this behavior fine for you guys?

dhhyi:
@M-Behr Yes, the main goal for us is that errors are mapped and available, so if that part of the PR is okay, we are good. We have customized the display in our project using toasts 😉 
@SGrueber Will you take care of the behavior for the standard PWA?

cc @ivopereira15

SGrueber:
Yes, I will

SGrueber:
The error display in the mini basket looks like this now:

image

MaxKless:
Hi @ivopereira15, can you let me know why you used the `info` basket substate to track the errors & causes? Instead of the `error` substate?
It would seem a bit cleaner that way but maybe that comes with some additional disadvantages. Could you provide some context?

M-Behr (approved):

ivopereira15:
Hi all. @MaxKless 

The error substate its working only when you have a bad request. Like a Status Code: 422 Unprocessable Entity. This is a clear error message that will be handled by `addItemsToBasketFail`

Lets imagine in the quickorder (that was one the problems that we found) we are trying to add to cart two products. One that we will have an error and other will be success. The response is Status Code: 207 Multi-Status. 
That means we are not mapping to an error. We are going to the`addItemsToBasketSuccess` action and mapping to an info object.  

here what we receive from the request.
image


What i did was put that error inside the infos object. Since we already were showing alert/info error using the info object, could be a good idea to show the error too.
With that infos object we have the info and the errors on the same place.

This works fine for us but might was not the best approach.

src/app/core/facades/checkout.facade.ts
```diff
@@ -85,6 +85,7 @@ export class CheckoutFacade {
   basketChange$ = this.basketChangeInternal$.asObservable();
   basketError$ = this.store.pipe(select(getBasketError));
   basketInfo$ = this.store.pipe(select(getBasketInfo));
+  basketInfoError$ = this.basketInfo$.pipe(map(info => (info?.length ? info[0].error : undefined)));
```
MaxKless, line 88: Maybe this should be its own selector?

Also, maybe this is the wrong place to ask this so please let me know if that's the case but can you explain the store structure here a bit? @SGrueber 
image

How come there's both a `basket.info` as well as a `info` substate in the basket state?

MaxKless (changes requested):

src/app/core/facades/checkout.facade.ts
SGrueber: basket.info contains info of the get basket (or the result of any call that returns the whole basket ). The info sub state contains the info of the last basket operation, e.g. addProductToCart. Because after each basket operation a get basket is triggered we need both info values.

SGrueber:
has been merged by #1230

PR #1115, opened 2022-04-14
URL encode formParamsToString values

JonaPaulus:
## PR Type

[x] Bugfix

When the filters coming from ICM are configured with an amp (&), there is a good chance that the product filters do not work - for example in the ish-filter-collapsible component.

Example:
Clicking a filter with e.g. the configured name "Laptops & Convertibles" would end up with a newly added filter option on the bottom of the list named "Laptops" that is selected instead of the correct "Laptops & Convertibles".

No product matches the filter.

Issue Number: --

## What Is the New Behavior?

This unexpected behavior is prevented by URL encoding the value.
So the amp (&) from the string does not appear in the filter parameters URL.

A unit test has been added

[ ] Yes
[x] No

## Other Information

--


shauke (approved):
Nice fix, thank you.

PR #1117, opened 2022-04-14
feat: add integration environment configuration

carlosDjangoo:
## PR Type

[ ] Bugfix
[x] Feature
[x] Build-related changes
[ ] CI-related changes
[ ] Other: 

There is only one configuration 'production' for all Server Environments. You can not decide to activate the Store/State Management or Source Map on integration test servers.

Issue Number: Closes #

## What Is the New Behavior?

There is another configuration 'integration' which will be triggered if the branch 'integration' or 'develop' will be deployed. Now you can configure a different configuration for the test systems.

The breaking change is that a deployment from the branch 'integration' or 'develop' will trigger the new configuration 'integration'.

[x] Yes
[ ] No

## Other Information

[AB#76041](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76041)

### DIFF
--- a/.azure/azure-pipelines.yml
+++ b/.azure/azure-pipelines.yml
@@ -55,7 +55,7 @@ stages:
             value: $(imageBaseName)-ssr
         steps:
           - bash: |
-              docker build -f $(system.defaultWorkingDirectory)/Dockerfile -t $(registryServerName)/$(imageName):$(tag) $(system.defaultWorkingDirectory) --build-arg displayVersion=$(tag)
+              docker build -f $(system.defaultWorkingDirectory)/Dockerfile -t $(registryServerName)/$(imageName):$(tag) $(system.defaultWorkingDirectory) --build-arg displayVersion=$(tag) --build-arg branchName=$(Build.SourceBranchName)
               docker login $(registryServerName) -u $(registryLogin) -p $(registryPassword)
               docker push $(registryServerName)/$(imageName):$(tag)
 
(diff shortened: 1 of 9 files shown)


dhhyi:
@carlosDjangoo 

That looks like a valid project customization that you can very well do in the project but should not be part of the standard (especially since it pins down branch names that can change from project to project depending on how the implementor sets it up).

It looks to me that you want to run a different build on test servers than you are actually pushing to production later. That is really risky since the production configuration optimizes things and in rare cases causes issues.

The way we have set it up in our project (with the Intershop Platform offering) is that we don't rebuild anything once the first image hits the deployments. After that we test exactly what will be going live on INT, re-tag the same docker image for UAT and test it there and finally instruct Intershop Support to deploy the exact same image to PRD.

Activating/Disabling sourceMaps to show on one environment and not in the other can be implemented as a runtime setting on the container so that doesn't need an additional build, I can look into that.
What do you mean with?
> decide to activate the Store/State Management

dhhyi:
@carlosDjangoo please have a look at #1127. Did you by any chance mean to enable the store dev tools? Just remove the part with the [file replacement](https://github.com/intershop/intershop-pwa/blob/2a6457fdd84f452bb2ef434b4310525b4b32e206/angular.json#L76-L81) in your project. It is perfectly okay to have store dev tools active on production environments, as it doesn't expose code to the user. On production systems the devtools are also run in logOnly mode.

PR #1122, opened 2022-04-15
refactor: delegate types in lazy-component schematic

dhhyi:
## PR Type

- Type updates in components which have the `@GenerateLazyComponent` decorator require a run of `npm run synchronize-lazy-components` to be reflected in the generated lazy components, because the initialization is copied.
- Generated lazy components contain eslint ignores, because the typing is not properly generated

## What Is the New Behavior?

- refactored lazy-component schematic to import the type of the original component and use it to delegate the typing of inputs
- improved readability of schematics template

[ ] Yes
[x] No

## Other Information

It is important that the original component is imported twice in the generated lazy component:
- the typing is imported with `import type` and therefor exists only in Typescript
- the lazy import with `await import` is used for webpack lazy loading

If the component would be imported with just `import` in the header, then the original component becomes part of the initial bundle which is something we don't want.

[AB#76046](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76046)

### DIFF
--- a/schematics/src/lazy-component/factory.ts
+++ b/schematics/src/lazy-component/factory.ts
@@ -65,8 +65,7 @@ export function createLazyComponent(options: Options): Rule {
     options = findDeclaringModule(host, options);
     options = determineArtifactName('component', host, options);
 
-    let bindings: { declaration: string; name: string }[] = [];
-    let imports: { types: string[]; from: string }[] = [];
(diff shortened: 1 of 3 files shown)


MaxKless (approved):
importing the component type and delegating is really a nice solution

PR #1134, opened 2022-04-20
refactor: access to checkoutpages handled by guard

paramaeder:
## PR Type

[ ] Bugfix
[ ] Feature
[X ] Refactoring (no functional changes, no API changes)
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently you can navigate to the checkout with a non-existent shopping cart via the address bar of the browser

## What Is the New Behavior?
Now the route is protected by a guard. Navigate to /checkout or any related childRoute without an existing basket will redirect you to the root url

[ ] Yes
[X ] No

### DIFF
--- a/src/app/pages/checkout/checkout-page.guard.ts
+++ b/src/app/pages/checkout/checkout-page.guard.ts
@@ -1,9 +1,9 @@
 import { Injectable } from '@angular/core';
 import { CanActivate, Router, UrlTree } from '@angular/router';
-import { Observable, debounceTime, map, of, switchMap, withLatestFrom } from 'rxjs';
+import { Observable, map, race, timer } from 'rxjs';
 
 import { CheckoutFacade } from 'ish-core/facades/checkout.facade';
-import { whenFalsy } from 'ish-core/utils/operators';
+import { whenTruthy } from 'ish-core/utils/operators';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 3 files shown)


src/app/pages/checkout/checkout.guard.ts
```diff
@@ -0,0 +1,25 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class CheckoutGuard implements CanActivate {
+  constructor(private checkoutFacade: CheckoutFacade, private router: Router) {}
+
+  canActivate(): boolean | UrlTree {
+    let hasBasket: boolean;
+
+    this.checkoutFacade.basket$.subscribe(value => {
```
MaxKless, line 15: I have to admit I'm a little surprised this works.  
Subscribing and then comparing the value extracted is not guaranteed to work - imagine if the `basket$` emission is a little delayed for whatever reason. In that case `hasBasket` will not be set correctly even though there might actually be a basket.

In addition, triggering a subscribe on method call without any logic to end the subscription is dangerous. Imagine if I tried to access the checkout route over and over again - every time `canActivate` is called, a new subscription will be opened. There are ways to fix this - check out an example like the [`feature-toggle.directive.ts`](https://github.com/intershop/intershop-pwa/blob/develop/src/app/core/directives/feature-toggle.directive.ts) to see how open subscriptions are canceled. However, this should only really be done in edge cases where nothing else is possible.
[Don't listen to me, listen to the rxjs team lead.](https://benlesh.medium.com/rxjs-dont-unsubscribe-6753ed4fda87).

Thankfully, there is a super simple solution here: [guards can return Observables](https://angular.io/api/router/CanActivate). So just return `this.checkoutFacade.basket$.pipe(....)` and do the value comparison in there.
Angular will take care of subscribing to the observable and using the emitted value for routing - nothing required on your part.

Also, I'm not sure if you need to return `this.router.parseUrl('/')` or if returning `false` would be sufficient.... Maybe you can check.

MaxKless (changes requested):

src/app/pages/checkout/checkout-page.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CheckoutPageGuard implements CanActivate {
+  constructor(private checkoutFacade: CheckoutFacade, private router: Router) {}
+
+  canActivate(): Observable<boolean | UrlTree> {
+    return this.checkoutFacade.basket$.pipe(
+      debounceTime(500),
```
Eisie96, line 16: Why is the debounceTime needed? I think all possible scenearios, that a basket is loaded or is currently loading, are checked in the following switchMap. And this rxjs operator leads to the problem, that the checkout route is not activated for the first 500ms, although a basket could be available.

SGrueber (changes requested):

src/app/pages/checkout/checkout-page.guard.ts
paramaeder: No debounceTime needed anymore. Guard now handles the request with a race operator

src/app/pages/checkout/checkout.guard.ts
paramaeder: Thanks Max! 
With increasing knowledge the original implementation was now completely overhauled

Eisie96 (approved):

MaxKless (approved):

MaxKless (approved):

PR #1142, opened 2022-05-02
chore: remove useless schematics vscode extension

MaxKless:
The Angular Schematics extension does not support the "schematics" option anymore, probably due to monetization. 
I removed it until an alternative is found.

[AB#76390](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76390)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -17,11 +17,9 @@
   "extensions": [
     // general
     "eamodio.gitlens",
-    "coenraads.bracket-pair-colorizer",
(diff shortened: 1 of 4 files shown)


shauke:
There is probably more that needs to be changed/removed in this context. Search for `cyrilletuzi.angular-schematics`.

shauke:
The functionality to work with our project specific schematics is available up to version 5.1.5. You can install a specific version of an extension by selecting "Install Another Version..." from the "Uninstall" selection button.

PR #1146, opened 2022-05-04
fix(schematics): make override schematic robust to path seperators

MaxKless:
Currently, the override schematic only accepts paths with forward slashes.

Now, it also accepts paths with backslashes.

[AB#76433](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76433)

### DIFF
--- a/schematics/src/helpers/override/factory.ts
+++ b/schematics/src/helpers/override/factory.ts
@@ -1,6 +1,6 @@
 import { Rule, SchematicsException, chain } from '@angular-devkit/schematics';
 import { getWorkspace } from '@schematics/angular/utility/workspace';
-import { normalize, sep } from 'path';
+import { normalize } from 'path';
 import { OverrideOptionsSchema as Options } from 'schemas/helpers/override/schema';
 
 import { copyFile } from '../../utils/filesystem';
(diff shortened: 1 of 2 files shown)


schematics/src/helpers/override/factory.ts
```diff
@@ -19,9 +20,15 @@ export function override(options: Options): Rule {
     const workspace = await getWorkspace(host);
     const project = workspace.projects.get(options.project);
     const sourceRoot = project.sourceRoot;
-    const from = `${
-      options.path ? `${options.path}/` : !options.from?.startsWith(`${sourceRoot}/app/`) ? `${sourceRoot}/app/` : ''
-    }${options.from.replace(/\/$/, '')}`;
+    const from = normalize(
+      `${
+        options.path
```
dhhyi, line 25: I would not use path sep and stick to slashes only.
just do `normalize(options.path).replace(/\\/, '/')` and keep the remaining.

schematics/src/helpers/override/factory_spec.ts
```diff
@@ -282,4 +282,19 @@ describe('override Schematic', () => {
       ]
     `);
   });
+
+  it('should override file if path contains backslashes', async () => {
+    const tree = await runOverride({
+      from: 'core\\routing\\product.route.ts',
```
dhhyi, line 288: On win i always have paths also starting with `.\`.

dhhyi (changes requested):

schematics/src/helpers/override/factory.ts
```diff
@@ -20,15 +20,14 @@ export function override(options: Options): Rule {
...
-        options.path
-          ? `${options.path}${sep}`
-          : !options.from?.startsWith(`${sourceRoot}${sep}app${sep}`)
-          ? `${sourceRoot}${sep}app${sep}`
-          : ''
-      }${options.from.replace(/\/$/, '')}`
-    );
+    const path = normalize(options.path ?? '')
+      .replace(/\\/g, '/')
+      .replace(/^\./, '');
```
dhhyi, line 25: `normalize` _should_ already take care of that.

dhhyi (changes requested):

schematics/src/helpers/override/factory_spec.ts
MaxKless: also works with those paths now.

schematics/src/helpers/override/factory.ts
dhhyi: @MaxKless Am I wrong here?

PR #1153, opened 2022-05-05
fix: only update cost center when it actually changes

MaxKless:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently, when you navigate to the basket page and the user has only one cost center, the cost center is nevertheless reassigned and because of this, the basket is reloaded. This leads to certain error/information messages dissapearing.

## What Is the New Behavior?
Now, the cost center is only reassigned when it actually differs from the one already set in the basket.

[ ] Yes
[x] No

## Other Information

[AB#76467](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76467)

### DIFF
--- a/src/app/shared/components/basket/basket-cost-center-selection/basket-cost-center-selection.component.ts
+++ b/src/app/shared/components/basket/basket-cost-center-selection/basket-cost-center-selection.component.ts
@@ -2,7 +2,7 @@ import { ChangeDetectionStrategy, ChangeDetectorRef, Component, OnDestroy, OnIni
 import { FormGroup } from '@angular/forms';
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { Observable, Subject } from 'rxjs';
-import { distinctUntilChanged, map, shareReplay, switchMap, take, takeUntil } from 'rxjs/operators';
+import { distinctUntilChanged, map, switchMap, take, takeUntil, withLatestFrom } from 'rxjs/operators';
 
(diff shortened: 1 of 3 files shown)


src/app/shared/components/basket/basket-cost-center-selection/basket-cost-center-selection.component.ts
```diff
@@ -71,11 +55,25 @@ export class BasketCostCenterSelectionComponent implements OnInit, OnDestroy {
         whenTruthy(),
         map(val => val.costCenter),
         distinctUntilChanged(),
+        withLatestFrom(this.checkoutFacade.basket$),
         takeUntil(this.destroy$)
       )
-      .subscribe(costCenter => {
-        this.checkoutFacade.updateBasketCostCenter(costCenter);
+      .subscribe(([costCenter, basket]) => {
+        if (costCenter !== basket.costCenter) {
```
SGrueber, line 62: This condition has to be a bit more precise - "", undefined and null values have to be treated as the same, if a user has more than one cost center 2 requests are send if the user enters cart page setting the cost center null and "" , but actually no request should be sent in this case, see image

SGrueber (changes requested):

SGrueber:
image

SGrueber:
image

M-Behr (approved):

SGrueber (approved):

PR #1154, opened 2022-05-10
chore: remove unused http-loader dep

MaxKless:
running `npx depcheck` showed an unused dependency, which is a leftover from one of the many translation reworks.
I removed it.


dhhyi:
@MaxKless we actually still use it in our project together with #927 😄 
Maybe I should instead finalize that PR to be included in the standard?

MaxKless:
Yes, that seems like a good idea. I'm definitely against keeping deps for projects but if that PR makes it into the standard, we can keep it around for now.

shauke (approved):

shauke (approved):

PR #1156, opened 2022-05-10
feat: use new token api service

Eisie96:
## PR Type

[ ] Bugfix
[ x ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The old token behavior with the token exchange for each request, should be replaced by the new token api service.

Issue Number: Closes #

## What Is the New Behavior?

The new token service should be used to authenticate a user. In case the user doesn't login to the PWA, the apiToken for an anonymous user should be used. The apiToken should be automatically refreshed before the token expires.

[x] Yes
[ ] No

## Other Information

- [x] Use new token service to get tokens for an anonymous user
- [x] Use new token service to get tokens for a user with credentials
- [x] Restore all necessary store informations on reload
- [x] Refresh before it is expired

To be discussed:
- further docu necessary ?
- when will the token expire if the user is inactive for a long time?

[AB#76575](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76575)


SGrueber (changes requested):
migration note is missing

SGrueber (changes requested):
xxx

src/app/extensions/punchout/identity-provider/punchout-identity-provider.spec.ts
```diff
@@ -67,10 +66,9 @@ describe('Punchout Identity Provider', () => {
   });
 
   beforeEach(() => {
-    cookieVanishes$ = new Subject<ApiTokenCookieType>();
     when(apiTokenService.restore$(anything())).thenReturn(of(true));
     when(checkoutFacade.basket$).thenReturn(EMPTY);
-    when(apiTokenService.cookieVanishes$).thenReturn(cookieVanishes$);
+    when(apiTokenService.cookieVanishes$).thenReturn(new Subject());
```
MaxKless, line 71: If it's used in the service, shouldn't `cookieVanishes$` be tested somehow?

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -72,32 +83,35 @@ export class ApiTokenService {
...
             } else if (orderId) {
               return { apiToken, type: 'order', orderId, creator: 'pwa' };
-            } else {
-              const apiTokenCookieString = this.cookiesService.get('apiToken');
-              const apiTokenCookie: ApiTokenCookie = apiTokenCookieString
-                ? JSON.parse(apiTokenCookieString)
-                : undefined;
-              if (apiToken && apiTokenCookie) {
+            }
+            const apiTokenCookieString = this.cookiesService.get('apiToken');
```
MaxKless, line 87: I think an explanatory comment could be useful here.

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -72,32 +83,35 @@ export class ApiTokenService {
...
           const cookieContent = apiToken?.apiToken ? JSON.stringify(apiToken) : undefined;
           if (cookieContent) {
-            cookiesService.put('apiToken', cookieContent, {
-              expires: new Date(Date.now() + 3600000),
-              secure: true,
-              sameSite: 'Strict',
-            });
+            if (!isEqual(apiToken, initialApiToken)) {
+              cookiesService.put('apiToken', cookieContent, {
+                expires: apiToken.options?.expires ?? new Date(Date.now() + 3600000),
```
MaxKless, line 104: Should the expiry date be configurable somehow? Maybe via an injection token or a static/normal property?
It would be easier to change if it weren't hard-coded here.

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -258,6 +262,16 @@ export class ApiTokenService {
     return this.appendAuthentication(req).pipe(
       concatMap(request =>
         next.handle(request).pipe(
+          map(event => {
+            if (event instanceof HttpResponse && event.url.endsWith('-/token') && request.body instanceof HttpParams) {
+              // eslint-disable-next-line @typescript-eslint/naming-convention
+              const { id_token, ...body } = event.body;
```
MaxKless, line 268: Could you not rename the destructured variable like this: `const { id_token: _, ...` and avoid the ESLint exception?

src/app/core/services/user/user.service.ts
```diff
@@ -78,29 +75,42 @@ export class UserService {
...
+    grantType: T,
+    options?: R
+  ): Observable<TokenResponse> {
+    return from(
+      this.oauthService.fetchTokenUsingGrant(
+        grantType,
+        options ?? {},
+        new HttpHeaders({ 'content-type': 'application/x-www-form-urlencoded' })
+      )
+    ).pipe(tap(tokenResponse => this.apiTokenService.setApiToken(tokenResponse.access_token)));
```
MaxKless, line 111: I think this setting logic should be handled in the effect, don't you?
Because usually a `fetch` method doesn't involve any writing and it can become confusing.

For example, the `fetchAnonymousUser` effect simply calls this method and does nothing with the return value -> it would be more transparent if it wrote the return value to the `apiTokenService` itself, imo.

Also, you could save a dependency on the `ApiTokenService` here, making everything less entangled.

src/app/core/identity-provider.module.ts
```diff
@@ -22,6 +22,7 @@ export function storageFactory(): OAuthStorage {
 
 @NgModule({
   imports: [OAuthModule.forRoot({ resourceServer: { sendAccessToken: false } }), PunchoutIdentityProviderModule],
+  /* eslint-disable @angular-eslint/sort-ngmodule-metadata-arrays */
```
jometzner, line 25: What was the purpose of this?

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -72,32 +83,35 @@ export class ApiTokenService {
...
           } else {
             cookiesService.remove('apiToken');
           }
         });
 
-      // token vanishes routine
+      // access token vanishes routine
       appRef.isStable
         .pipe(
           whenTruthy(),
```
jometzner, line 117: Instead of `interval(1000)` did you consider using `cookiesOnChanged()` ?

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged

src/app/core/utils/api-token/api-token.service.ts
jometzner: Zu Befehl

src/app/core/utils/api-token/api-token.service.ts
MaxKless: This is an add-on feature and not actually supported by default.

jometzner:
If I configure PWA to use some none-sense `identityProvider:` value, I get a very weird error. Maybe we should include some well articulated error message?

```
default-error-handler.ts:18 TypeError: Cannot read properties of undefined (reading 'intercept')
    at IdentityProviderInterceptor.intercept (identity-provider.interceptor.ts:15:56)
    at HttpInterceptorHandler.handle (http.mjs:1442:33)
    at ICMErrorMapperInterceptor.intercept (icm-error-mapper.interceptor.ts:100:17)
    at HttpInterceptorHandler.handle (http.mjs:1442:33)
    at DefaultOAuthInterceptor.intercept (angular-oauth2-oidc.mjs:3127:18)
    at HttpInterceptorHandler.handle (http.mjs:1442:33)
    at HttpXsrfInterceptor.intercept (http.mjs:2048:25)
    at HttpInterceptorHandler.handle (http.mjs:1442:33)
    at HttpInterceptingHandler.handle (http.mjs:2094:27)
    at http.mjs:1269:70
```

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -48,8 +55,12 @@ export class ApiTokenService {
     private cookiesService: CookiesService,
     private router: Router,
     private store: Store,
+    oAuthService: OAuthService,
+    oAuthConfigurationService: OAuthConfigurationService,
     appRef: ApplicationRef
   ) {
+    oAuthConfigurationService.config$.subscribe(config => oAuthService.configure(config));
```
jometzner, line 62: If I run this with a configured `type: 'auth0'` identity provider, our `OAuthService` lib cannot get the initial discovery document loaded. It seems as though the different closure contexts are mixed up. One `oAuthService` uses ICM token endpoint, another is using whatever is configured with the `type: 'auth0'` client.

jometzner (changes requested):

src/app/core/identity-provider.module.ts
Eisie96: This is not necessary anymore.

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -48,8 +55,12 @@ export class ApiTokenService {
...
+
     const initialCookie = this.parseCookie();
```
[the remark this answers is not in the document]
Eisie96: I think so. The oauth library should only handle the management of the api-token. Nevertheless the PWA needs to store some other informations (is user anonymous, has user a basket or an order) to restore informations on reload. This could be done with the initial cookie handling. It can be discussed, whether the apiToken should be part of the cookie or not.

jometzner (approved):

src/app/core/utils/oauth-configuration/oauth-configuration.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const TOKEN_ENDPOINT = '-/token';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OAuthConfigurationService {
+  constructor(private apiService: ApiService) {}
+
+  get config$(): Observable<AuthConfig> {
```
Eisie96, line 16: Configuration is set by deployment settings and is not dynamic.

src/app/core/utils/api-token/api-token.service.ts
jometzner: Or a test

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -88,16 +108,16 @@ export class ApiTokenService {
           const cookieContent = apiToken?.apiToken ? JSON.stringify(apiToken) : undefined;
           if (cookieContent) {
             cookiesService.put('apiToken', cookieContent, {
-              expires: new Date(Date.now() + 3600000),
-              secure: true,
+              expires: apiToken.options?.expires ?? new Date(Date.now() + ApiTokenService.DEFAULT_EXPIRY_TIME),
```
jometzner, line 111: As far as I can see `apiToken` object doesn't contain `options` key. How do you map the `expires_in` and `refresh_expires_in` characteristics of the token resource object?

jometzner:
I occasionally get this message logged in my console.
```
default-error-handler.ts:18 Error: 'tokenEndpoint' should not be null
    at Y.assertUrlNotNullAndCorrectProtocol (angular-oauth2-oidc.mjs:1220:19)
    at Y.fetchTokenUsingGrant (angular-oauth2-oidc.mjs:1554:14)
    at _.fetchToken (user.service.ts:100:25)
    at user.effects.ts:126:42
    at z.subscribe.G (switchMap.js:14:23)
    at _._next (OperatorSubscriber.js:13:21)
    at _.next (Subscriber.js:31:18)
    at filter.js:6:128
    at _._next (OperatorSubscriber.js:13:21)
    at _.next (Subscriber.js:31:18)
```
Seems like you left some spots in which the current configuration is not yet ready.

src/app/core/store/customer/user/user.actions.ts
```diff
@@ -25,6 +25,12 @@ export const loadCompanyUserSuccess = createAction('[User API] Load Company User
 
 export const logoutUser = createAction('[User] Logout User');
 
+export const revokeApiToken = createAction('[User] Revoke Api Token');
```
SGrueber, line 28: introduce logoutUserSuccess and logoutUserFail actions,  use [User API] in brackets

remove revokeApiToken actions

src/app/core/utils/api-token/api-token.service.ts
```diff
@@ -39,9 +59,14 @@ interface ApiTokenCookie {
 
 @Injectable({ providedIn: 'root' })
 export class ApiTokenService {
+  // If no expiry date is supplied by the token endpoint, this value (in ms) is used
+  private static DEFAULT_EXPIRY_TIME = 3600000;
```
SGrueber, line 63: declare this variable as const on top of this service

SGrueber (changes requested):
pls add a short migration note

pls add an authentification documentation (concept) and explain
what happens
 when the pwa is started (token, cookie)
- when the user logs in
- when the user logs out
- when the token expires
- what the oauth.service is used for
- and all other stuff you think it is important to know

SGrueber (changes requested):

SGrueber (dismissed):
The routes /invite and /forgotPassword/updatePassword don't work any more in case of SSO - there is an endless loading animation shown

shauke (approved):

shauke (dismissed):

mglatter:
Further adjustments of the other .md files may follow after having reviewed them as well.

mglatter:
I found some more things I could not comment on directly in the files as no changes have been made in the lines in question:

migrations.md

line 94 says: "The [@ngx-translate/http-loader](https://github.com/ngx-translate/core) dependency was removed since we did not use it."
The link text says ngx-translate/http-loader but the link opens ngx-translate/core. Is this correct?


nginx-startup.md

Line 20 says: "- Connect **it** to the PWA with `UPSTREAM_PWA` in the form of `http://<IP>:<PORT>`"
Can you please specify and replace "it" with the concrete name of what should be connected? Else it is not clear if nginx or a Docker image is meant to be connected (at least to me not being into the topic deeply). ;-) Thanks

shauke:
* I checked and the link is correct.
* fixed the sentence and used "nginx" for "it"

shauke (approved):

PR #1157, opened 2022-05-10
feat: select displays fixed string for single option required fields

MaxKless:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently, required select option fields with only one option will still show a (unnecessary) select box.

Issue Number: Closes #

## What Is the New Behavior?
Now, the select option field type shows a fixed string if there is only one (non-placeholder) option and it's required

## ToDo:
- [ ] add `:` after fixed display

[ ] Yes
[x] No

## Other Information

[AB#76576](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76576)

### DIFF
--- a/src/app/shared/formly/types/select-field/select-field.component.html
+++ b/src/app/shared/formly/types/select-field/select-field.component.html
@@ -1,18 +1,24 @@
-<select
-  [formControl]="formControl"
-  [compareWith]="to.compareWith ? to.compareWith : defaultOptions.templateOptions.compareWith"
-  [formlyAttributes]="field"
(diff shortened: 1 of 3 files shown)


M-Behr (approved):

SGrueber:
After involvement of IAD/VD it is still not clear that the new behavior is alsways better than the exisiting one. That's why we decided to keep the select boxes so far even if there is only one option in it.

PR #1162, opened 2022-05-17
fix: lazy check for available product links

dhhyi:
## PR Type

[x] Bugfix

If the property `displayOnlyAvailableProducts` is activated on `ish-product-links-carousel`, the component fetches all linked products to calculate the carousel content. This leads to a drop in page performance, especially in SSR if products have many linked accessories.

## What Is the New Behavior?

The product links are evaluated lazy slide by slide whenever the user switches to them. For SSR only the first slide is evaluated.

[ ] Yes
[x] No

## Other Information

[AB#76738](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76738)

### DIFF
--- a/src/app/pages/product/product-links-carousel/product-links-carousel.component.html
+++ b/src/app/pages/product/product-links-carousel/product-links-carousel.component.html
@@ -6,7 +6,7 @@
       <swiper [config]="swiperConfig">
         <ng-template *ngFor="let sku$ of productSKUs" swiperSlide let-data>
           <ish-product-item
-            *ngIf="data.isVisible && sku$ | async as sku"
+            *ngIf="lazyFetch(data.isVisible, sku$) | async as sku"
             ishProductContext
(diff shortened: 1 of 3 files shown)


SGrueber:
@dhhyi 
I linked 7 accessory products to a product, one of them is not available.
If I scroll to the left and after that to the right so products become visible again I experience the following issues: 
- the image of these products are not visible immediately so that the no-image-available  image is to be seen shortly
- the product prices for these products are (unnecessarily) fetched from the REST api

Is there any chance to fix these issues?

dhhyi:
@SGrueber 
> Is there any chance to fix these issues?

Yes, this behavior was caused by tearing down and re-initializing product contexts. This is now fixed by preserving the slides that have already been viewed.

SGrueber (approved):

PR #1163, opened 2022-05-17
feat(seo): optimized category URLs with full category path

Eisie96:
## PR Type

[ ] Bugfix
[ ] Feature
[ x ] Refactoring (no functional changes, no API changes)
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

## What Is the Behavior?

The current category url contains only the current category displayname with an appended categoryUniqueId. (/Acer-Sub-catComputers.897.897_Acer.Acer-Sub)

This needs to be improved for SEO. Every category page should contain the whole path from the top level to the requested category. Only display-names should be used. The needed categoryUniqueId to get the data from the ICM should be at the end of the url. (/acer/sub/acer-sub-catComputers.897.897_Acer.Acer-Sub)

[x] Yes

## Other Information

[AB#76756](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/76756)


src/app/core/store/shopping/categories/categories.selectors.ts
```diff
@@ -73,8 +90,20 @@ export const getNavigationCategories = (uniqueId: string) =>
     defaultMemoize(projector, CategoryTreeHelper.equals, isEqual)
   )(getCategoryTree, (tree: CategoryTree): NavigationCategory[] => {
     if (!uniqueId) {
-      return tree.rootIds.map(mapNavigationCategoryFromId.bind(tree));
+      return tree.rootIds.map(id =>
+        mapNavigationCategoryFromId.bind({
```
MaxKless, line 94: Maybe this isn't your code originally, but I think the `bind` construct is a bit weird. 
Maybe we can make it simpler?

src/app/core/routing/category/category.route.ts
```diff
@@ -13,7 +13,8 @@ export function generateLocalizedCategorySlug(category: Category) {
       ?.replace(reservedCharactersRegEx, '-')
       .replace(/-+/g, '-')
       .replace(/-+$/, '')
-      .replace('-cat', '-Cat') || ''
+      .replace('-cat', '-Cat')
```
MaxKless, line 16: replacement is unnecessary because everything is lowercased afterwards.

src/app/extensions/seo/store/seo/seo.effects.ts
```diff
@@ -64,7 +64,12 @@ export class SeoEffects {
             // PRODUCT PAGE
             this.productPage$.pipe(map(product => this.baseURL + generateProductUrl(product).substring(1))),
             // CATEGORY / FAMILY PAGE
-            this.categoryPage$.pipe(map(category => this.baseURL + generateCategoryUrl(category).substring(1))),
+            this.categoryPage$.pipe(
+              mergeMap(category =>
```
MaxKless, line 68: why `mergeMap` instead of `switchMap`?
It seems like outdated category information could safely be thrown away.

src/app/core/routing/category/category-route.pipe.ts
```diff
@@ -1,12 +1,30 @@
...
-export class CategoryRoutePipe implements PipeTransform {
+export class CategoryRoutePipe implements PipeTransform, OnDestroy {
+  private pathElements: Category[];
+
+  private destroy$ = new Subject<void>();
+
+  constructor(private shoppingFacade: ShoppingFacade) {}
   transform(category: CategoryView): string {
-    return generateCategoryUrl(category);
+    this.shoppingFacade
```
MaxKless, line 18: I don't like this. First of all this will probably create many subscriptions that are never disposed of. 
Second, there's no guarantee that the right `pathElements` will be used in the return. 

Either:
- return an observable and chain with an `async` pipe. A bit verbose but clean
- there's probably some other solution using impure pipes that keeps track of the category state and marks itself for change detection when the categoryPathElements change. 

I would have to do some more research here but I'm sure there's a nice solution. 
Maybe something completely different exists too.

MaxKless (changes requested):

src/app/core/store/shopping/categories/categories.selectors.ts
MaxKless: Definitely not a blocker, but something we should talk about as a team. Conciseness vs. readability, I guess.

Eisie96: @dhhyi Can you help to understand the reason to use a binding instead of a function parameter

dhhyi: It is less verbose using bind. But yes, it is also not a commonly known feature. Feel free to change it to a function parameter implementation.

src/app/core/routing/category/category-route.pipe.ts
Eisie96: Yeah! Your definetly right. :) A good solution for that is documented in the offical Angular docs. (https://angular.io/guide/pipes#caching-http-requests) It uses a caching feature to subscribe only then, when the input (category) changes. In addition the custom pipe mark itself for a change detection whenever the categoryPathElements$ observable emits a new value.

MaxKless: Nice solution, but I have two more questions:
- Why do you mark for changes manually when you can use the `pure: false` attribute to do it for you (as shown in the docs)
- You still run into the issue of aggregating subscriptions. Since `categoryPathElements$` is a long running observable, unlike `httpClient.get`, you also need to unsubscribe from the old subscription whenever a new value is requested. Otherwise, you might have multiple long running subscriptions leaking.

dhhyi (changes requested):
@Eisie96 Is there anything that speaks against including the path categories into the `CategoryView` model? That way you wouldn't need an additional store selector and additional arguments to the functions... 🤔

Eisie96:
@dhhyi Thank you for your review. It is a really good idea to put the category path to the category view! The CategoryRoutePipe is now a pure function. One thing i have to change would be the getCategory(uniqueId) selector. It generates the category view with the subtree of the given uniqueid. But the whole tree is necessary to get all elements from the category path.

src/app/core/routing/category/category-route.pipe.ts
```diff
@@ -7,6 +7,10 @@ import { generateCategoryUrl } from './category.route';
 @Pipe({ name: 'ishCategoryRoute', pure: true })
 export class CategoryRoutePipe implements PipeTransform {
   transform(category: CategoryView): string {
-    return generateCategoryUrl(category);
+    if (category) {
+      return generateCategoryUrl(category);
+    }
+
+    return;
```
shauke, line 14: Why this additional check? This is done in the function already (and also done only in the function for product and page routes).

shauke (approved):

MaxKless (approved):

shauke (approved):

PR #1178, opened 2022-06-02
fix: window object should only be used on browser side

Eisie96:
## PR Type

If the tracking extension is enabled **AND for some reason the cookies consent check is removed (project customization to use a different cookies consent management tool)**, then the tracking-config.effect.ts will use the window object on server side if a gtmToken is found.

## What Is the New Behavior?

Making sure that the window object is only used on browser side independent from the potentially removed cookies consent handling.

[x] No

## Other Information

[AB#77211](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/77211)

### DIFF
--- a/src/app/extensions/tracking/store/tracking-config/tracking-config.effects.ts
+++ b/src/app/extensions/tracking/store/tracking-config/tracking-config.effects.ts
@@ -1,4 +1,4 @@
-import { isPlatformServer } from '@angular/common';
+import { isPlatformBrowser, isPlatformServer } from '@angular/common';
 import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
 import { Actions, createEffect } from '@ngrx/effects';
 import { Store, select } from '@ngrx/store';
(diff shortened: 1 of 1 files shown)


shauke (approved):

src/app/extensions/tracking/store/tracking-config/tracking-config.effects.ts
```diff
@@ -24,7 +24,7 @@ export class TrackingConfigEffects {
     store: Store,
     cookiesService: CookiesService
   ) {
-    if (cookiesService.cookieConsentFor('tracking')) {
+    if (cookiesService.cookieConsentFor('tracking') && isPlatformBrowser(this.platformId)) {
```
shauke, line 27: This is actually an obsolete additional check since a cookie consent will only return true `if (isPlatformBrowser(this.platformId)) {`.
The project that had SSR errors because of the missing `window` object removed the cookie condition without replacing it with a platform check. So it might be a helpful pointer even though it is not necessary?

shauke (changes requested):

shauke (approved):

src/app/extensions/tracking/store/tracking-config/tracking-config.effects.ts
shauke: Changed the order of the conditions to make the importance of being in the browser more obvious and more independent from the check done by the cookies consent.

PR #1183, opened 2022-06-08
Feature: send PGID in promotions API

cesarGamaSa:
## PR Type

[ ] Bugfix
[X] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

When getting a promotion's data (Promotions API call), the PGID is not sent in the header of the call.

## What Is the New Behavior?

The PGID is sent in the header of the call.

[ ] Yes
[X] No


SGrueber (changes requested):
If you fetch promotions with a pgid you have to make sure that promotions are removed from store whenever the user login status changes (login/logout). Hence, register the promotion substate for the resetSubStatesOnActionsMeta reducer in the shopping-store.module, please.

shauke:
@cesarGamaSa What is the reason for adding the PGID to this specific promotions details call? As far as the ICM REST API is concerned no different response will ever be returned. There seems to be no need for this additional information.

shauke:
Hi @cesarGamaSa can you elaborate on the issue in regards to the missing PGID?
As I mentioned before, we currently see no valid reason to add this additional complexity to the promotions data handling if we do not understand the reason for it. There cannot be different responses for the standard ICM since we cannot manage any user specific promotion details in the Backoffice.

PR #1184, opened 2022-06-08
chore: add ci option to commitizen config

MaxKless:
According to our docs, we follow conventional commits (which in turn follows the angular conventions).
Both allow a `ci` type for commit messages, but we don't.

Added a `ci` type.

[AB#77327](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/77327)


shauke (dismissed):
nice addition 😉

PR #1196, opened 2022-06-27
style: enable eslint rules 'no-var' and ' prefer-const'

shauke:
enable eslint rules 'no-var' and ' prefer-const' plus source code fixes that did not fit these rules

## PR Type

[x] Code style update (formatting, local variables)

[x] No

## Other Information

[AB#77715](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/77715)

### DIFF
--- a/src/typings.d.ts
+++ b/src/typings.d.ts
@@ -2,12 +2,12 @@ declare module 'express-http-proxy';
 
 declare module 'express-robots-txt';
 
-declare let PRODUCTION_MODE: boolean;
+declare const PRODUCTION_MODE: boolean;
 
-declare let SERVICE_WORKER: boolean;
+declare const SERVICE_WORKER: boolean;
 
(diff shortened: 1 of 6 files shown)


src/typings.d.ts
```diff
@@ -2,12 +2,12 @@ declare module 'express-http-proxy';
 
 declare module 'express-robots-txt';
 
-declare var PRODUCTION_MODE: boolean;
+declare let PRODUCTION_MODE: boolean;
```
dhhyi, line 5: In #1080 I changed these to const, which makes more sense in my opinion.

dhhyi (changes requested):

src/typings.d.ts
shauke: So, changed to 'const' which is even better. ;)
Thanks for the feedback.

dhhyi (approved):

PR #1202, opened 2022-07-07
chore: move xliff to dev dep

MaxKless:
The xliff dependency is only used in a single script but was a normal dependency. Now it's a dev dependency.


shauke (approved):
I approve this if you use `deps:` instead of `chore:` when merging into develop. 😉

PR #1208, opened 2022-07-11
deps: use a fixed node.js version to prevent build issues with unknown versions

shauke:
* fixed to node.js 16.16.0 including npm 8.11.0

## PR Type

[x] Build-related changes

docker and git actions use latest node.js 16 version that led to unexpected behavior

## What Is the New Behavior?

fixed node.js version does not introduce unexpected behavior/issues

[x] No

## Other Information

[AB#78109](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/78109)

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -9,6 +9,11 @@ kb_sync_latest_only
 
 ## 2.4 to 3.0
 
+With the 2.4.1 Hotfix we introduced a more fixed Node.js version handling to the version used and tested by us.
+We set Node.js 16.16.0 and npm 8.11.0 as our application runtime and package management versions.
+This is supposed to prevent unexpected build issues in the future but requires manual updating of Node.js to newer versions if tested successfully.
+Other Node.js versions might still work but you might get warnings regarding the projects recommended settings.
+
 The Intershop PWA 3.0 release includes a Jest Update to version 28, see also https://jestjs.io/docs/upgrading-to-jest28.
 The jest-marbles package has been replaced by jasmine-marbles.
 
--- a/.github/workflows/automated-upgrade.yml
+++ b/.github/workflows/automated-upgrade.yml
@@ -9,6 +9,9 @@ on:
   schedule:
     - cron: '0 0 * * 1'
 
+env:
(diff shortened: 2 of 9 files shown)


jometzner:
Maybe we could introduce some variable for the used nodejs version in GitHub pipelines? Changing the version number everywhere is not a pleasant task.

shauke:
I could either use a GitHub secret for the node version number, which is the easiest way but not very transparent since the actual value is hidden in the management view.

Or I could try to use this action to set a global variable that could be used by all workflows (https://github.com/marketplace/actions/set-environment-variables) but this is kind of much overhead for the problem at hand, but maybe just the one value to start with for more to come?

Or we could leave it to be changed everywhere. ;)  

What is the preferred action?

jometzner:
> I could either use a GitHub secret for the node version number, which is the easiest way but not very transparent since the actual value is hidden in the management view.
> 
> Or I could try to use this action to set a global variable that could be used by all workflows (https://github.com/marketplace/actions/set-environment-variables) but this is kind of much overhead for the problem at hand, but maybe just the one value to start with for more to come?
> 
> Or we could leave it to be changed everywhere. ;)
> 
> What is the preferred action?

Yes, I noticed that github has no such thing as variable groups that can be referred inside the workflow. I guess the best compromise is to use a environment variable. Scope it using the entire workflow like so:
```
on:
  bla

env:
  NODEJS_VERSION: 16.16.0

jobs:
```
I'll commit an example.

docs/guides/migrations.md
```diff
@@ -9,6 +9,11 @@ kb_sync_latest_only
 
 ## 2.4 to 3.0
 
+With the 2.4.1 Hotfix we introduced a more fixed Node.js version handling.
+For the docker containers and as the projects engine we set `Node.js 16.16.0` and `npm 8.11.0` to more specifically refer to the versions used and tested.
```
jometzner, line 13: Sorry I have trouble understanding that sentence. May I suggest:
`We set Node.js 16.16.0 and npm 8.11.0 as our application runtime and package management versions.`

PR #1213, opened 2022-07-14
feature: lazy load content for modal dialog link component

### LINKED ISSUE #1209: Performance: Content for ready-to shipment dialog is requested for each basket line item
**Actual Behavior**
Content for ready-to shipment dialog is requested for each basket line item. This might lead to performance problems in case of huge baskets.

**Expected Behavior**
The content is only loaded once per basket or if the user opens the dialog.

1. Add some products to cart and go to cart page
2. Check the network tab of your browser development tools

image

**Environment Details**

- Desktop: \*

  - Browser [e.g. chrome, safari], Version [e.g. 22]
  - Project Code Version

- Smartphone: \*
  - Browser [e.g. stock browser, safari], Version [e.g. 22]
  - Project Code Version

Eisie96:
## PR Type

[x] Feature

Each content include, which should be displayed in a modal dialog component, is rendered before modal is opened. This could lead to unnecessary REST calls.

Issue Number: Closes #1209

## What Is the New Behavior?

Content includes for modal dialog components can be lazy loaded, when the modal is opened.

[x] No

## Other Information

The content include needs to be marked with the LazyContentDirective.
The dialog component has now access to the template reference and can lazy load the input content, when the modal is opened.

```
<ish-modal-dialog-link linkText="checkout.tac.link" [options]="{ titleText: 'checkout.termsandconditions.details.title' | translate size: 'lg' }">
      <ish-content-include includeId="include.dialog.privacyPolicy.pagelet2-Include" *ishLazyLoadingContent></ish-content-include>
  </ish-modal-dialog-link>
```

[AB#78214](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/78214)

### DIFF
--- a/src/app/shared/components/common/modal-dialog-link/modal-dialog-link.component.ts
+++ b/src/app/shared/components/common/modal-dialog-link/modal-dialog-link.component.ts
@@ -10,7 +10,7 @@ import { ModalDialogComponent, ModalOptions } from 'ish-shared/components/common
  * The component is not designed or intended to contain any logic, but only to display text.
  *
  * The dynamic input content could be loaded on initialization (default) or on demand with lazy loading. This can be set via the lazyLoading input parameter.
- * If the content should be loaded on demand, then the input content must be wrapped within a ng-container, which uses the LazyLoadingContentDirective.
+ * If the content should be loaded on demand, then the input content must apply the structural directive LazyLoadingContentDirective.
  * The component has now access to the template reference and can lazy load the input content, when the modal is opened.
  *
  * @example
@@ -19,9 +19,7 @@ import { ModalDialogComponent, ModalOptions } from 'ish-shared/components/common
     [options]="{ titleText: 'checkout.termsandconditions.details.title' | translate, size: 'lg' }"
     [lazyLoading]="true"
   >
-    <ng-container *ishLazyLoadingContent>
-      <ish-content-include includeId="include.dialog.privacyPolicy.pagelet2-Include"></ish-content-include>
-    </ng-container>
+    <ish-content-include includeId="include.dialog.privacyPolicy.pagelet2-Include" *ishLazyLoadingContent></ish-content-include>
   </ish-modal-dialog-link>
  */
 @Component({
--- a/src/app/shared/components/common/modal-dialog-link/modal-dialog-link.component.ts
+++ b/src/app/shared/components/common/modal-dialog-link/modal-dialog-link.component.ts
@@ -1,4 +1,5 @@
 import { ChangeDetectionStrategy, Component, ContentChild, Input, ViewChild } from '@angular/core';
+import { Observable, last, map } from 'rxjs';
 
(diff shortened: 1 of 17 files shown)


src/app/shared/components/common/modal-dialog/modal-dialog.component.ts
```diff
@@ -67,6 +70,11 @@ export class ModalDialogComponent<T> {
     const size = this.options?.size ? this.options.size : undefined;
 
     this.ngbModalRef = this.ngbModal.open(this.modalDialogTemplate, { size });
+
+    this.shown$ = this.ngbModalRef.shown.pipe(
```
MaxKless, line 74: Is this used anywhere? Looks like dead code.

src/app/shared/components/common/modal-dialog-link/modal-dialog-link.component.ts
```diff
@@ -8,12 +9,19 @@ import { ModalDialogComponent, ModalOptions } from 'ish-shared/components/common
...
+ * The component has now access to the template reference and can lazy load the input content, when the modal is opened.
+ *
  * @example
  *<ish-modal-dialog-link
     linkText="checkout.tac.link"
     [options]="{ titleText: 'checkout.termsandconditions.details.title' | translate, size: 'lg' }"
+    [lazyLoading]="true"
   >
-    <ish-content-include includeId="include.dialog.privacyPolicy.pagelet2-Include"></ish-content-include>
+    <ng-container *ishLazyLoadingContent>
```
MaxKless, line 22: Since you use a structural directive that expands into an `<ng-template>`, are you sure you need to wrap the lazy rendered content in an `<ng-container>`? 

Have you tried simply applying the structural directive to the content?

MaxKless (changes requested):
One small fix and one small question.
Otherwise, a very clean solution - especially the part where you can switch between lazy and eager rendering with an input parameter.

src/app/shared/components/common/modal-dialog/modal-dialog.component.ts
Eisie96: It is used in the modal-dialog-link component as a condition for the content projection.

shauke:
To bad that it needs to be a combination of  `[lazyLoading]="true"` and `*ishLazyLoadingContent` and cannot be handled with one marker. Otherwise a nice unintrusive addition.
I added a few other places where this is useful too.
@Eisie96 A loading animation would not be possible as well? ;)

Eisie96:
@shauke I refactored the modal-link component, that the decision to load the content on demand depends if the LazyLoadingContentDirective is used for the content projection or not.

The loading spinner could be added as an element within the content projection. Nevertheless this is not possible, if the input content is a content-include. Until now there is no loading-flag for content-include entities available in the ngrx store.

MaxKless (dismissed):

shauke (dismissed):

PR #1231, opened 2022-08-01
chore(eslint): allow underscore variable names

MaxKless:
Currently, vars are allowed to be unused when they're named `_`. However, `_` is not an allowed variable name so there will be another ESLint exception isntead.

=> Fixed

[AB#78545](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/78545)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -99,7 +99,11 @@
           {
             "selector": "variable",
             "modifiers": ["const"],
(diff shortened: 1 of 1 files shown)


jometzner (approved):

SGrueber:
As far as I remember, this happened intentionally. We didn't want to support _ variables to indicate private variables. jm2c

PR #1250, opened 2022-08-18
fix: respect navigation root when displaying the breadcrumb of cms pages

SGrueber:
## PR Type

[x] Bugfix

The breadcrumb of cms pages contains always the whole path of the page tree even if a deferring navigation root is defined.

## What Is the New Behavior?

The display of the cms breadcrumb respects the defined navigation root.

[x] No

## Other Information

[AB#78918](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/78918)

### DIFF
--- a/src/app/core/facades/cms.facade.ts
+++ b/src/app/core/facades/cms.facade.ts
@@ -48,6 +48,12 @@ export class CMSFacade {
     return this.store.pipe(select(getContentPageTree(rootId)));
   }
 
+  /**
+   *
+   * @param rootId is taken into consideration as first element of breadcrumb for content page
+   *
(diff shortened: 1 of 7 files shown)


src/app/shared/cms/components/cms-static-page/cms-static-page.component.html
```diff
@@ -1,11 +1,14 @@
 <div class="content-page">
   <div class="row">
+    <ish-error-message [error]="contentPageTreeError$ | async"></ish-error-message>
```
shauke, line 3: Why would we need error handling here?
Maybe loading yes, but I am not sure about the error.

Eisie96 (dismissed):

PR #1254, opened 2022-08-23
fix: add missing translation key for company information on registration page

SGrueber:
## PR Type

[x] Bugfix

## What Is the New Behavior?

- The missing translation key for 'Company Information' has been added on registration page.

[x] No

## Other Information

[AB#79014](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79014)


dhhyi:
The `merge` was introduced to support filtering from product listing pages to master pages.
One of the use cases:
1. go to https://intershoppwa.azurewebsites.net/
2. search for `Microsoft`
3. Filter `Hard Disk Drive Capacity` to `1TB`
4. Click on either product tile
5. Master page has the same filters applied.

removing the merges will drop this functionality

It looks to me that those merges just have to be prevented for the recently functionality, because the products are not displayed within a filter context.

shauke (approved):

PR #1256, opened 2022-08-23
fix: do not overide product information in the state with information with lower completeness level

shauke:
## PR Type

[x] Bugfix

Product images and description are sometimes missing on the product detail page when routing there from a product list.

This happens if the category products call responds later then the product details call.
Both calls are triggered when routing to the product details page.

Problem 1: There is no need for the additional category products call when routing to the product details page.

Problem 2: Product information with a lower completeness level should not overwrite product information in the state with a higher completeness level.

## What Is the New Behavior?

The first commit addresses Problem 1 by reverting the commit that introduced this wrong behavior.

The second commit fixes the product information state update handling and considers the different completeness levels.
* update the complete product information in the state if the new product information has a higher or equal completeness level than the current product information in the state
* always update the availability information in the state with the new product information

[x] No

## Other Information

The fix for the issue of #1140 is reverted since it introduced more problems then it fixed. A new fix for #1140 will be necessary.

[AB#79022](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79022)

### DIFF
--- a/src/app/core/store/shopping/products/products.reducer.ts
+++ b/src/app/core/store/shopping/products/products.reducer.ts
@@ -2,7 +2,7 @@ import { EntityState, createEntityAdapter } from '@ngrx/entity';
 import { createReducer, on } from '@ngrx/store';
 
 import { ProductLinksDictionary } from 'ish-core/models/product-links/product-links.model';
-import { AllProductTypes, SkuQuantityType } from 'ish-core/models/product/product.model';
+import { AllProductTypes, ProductHelper, SkuQuantityType } from 'ish-core/models/product/product.model';
 
 import {
   loadProductFail,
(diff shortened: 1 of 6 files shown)


src/app/core/store/shopping/products/products.reducer.ts
```diff
@@ -50,13 +50,26 @@ export const productsReducer = createReducer(
...
+        failed: removeFailed(state.failed, product.sku),
+      });
     }
-
+    // always update the availability information in the state with the new product information
+    const newProduct = {
+      ...oldProduct,
+      available: product.available ?? oldProduct.available,
+      availableStock: product.availableStock ?? oldProduct.availableStock,
+    };
```
dhhyi, line 72: Please put the merging logic in a separate exported function and write tests for it.
From personal experience: This is probably customized in projects, so testing will help specifying custom behavior.

dhhyi: Hint: the merging function should accept two products and return the product that is upserted in the reducer. Two upsert statements with one in between the logic make all of this quite unreadable.

shauke: Thanks for the constructive critique. I hope it is now more the way you wished it would have been?
True, it looks better and I even wrote tests. 😄

SGrueber (approved):

PR #1265, opened 2022-09-08
Quoting feature results in a non stable application on certain pages

### LINKED ISSUE #1264: Quoting feature results in a non stable application on certain pages
**Actual Behavior**

If the 'quoting' feature is enabled, then all pages, where quoting components are rendered (/account, /account/quotes), are not stable.

**Expected Behavior**

All pages should be stable at an appropiate time.

1. Enable 'quoting' feature
2. Go to seo.effect.ts and add the "log('stable')" operator as first element to 'appRef.isStable' pipe.
3. Refresh on the account page (/account) with a logged in user and look in the console. (only false should be emitted)
4. Repeat 1-3 with 'quoting' feature disabled and the application should be stable at an appropiate time. (true should be emitted)

**Environment Details**

- Desktop: \*

  - Browser [e.g. chrome, safari], Version [e.g. 22]
  - Project Code Version

- Smartphone: \*
  - Browser [e.g. stock browser, safari], Version [e.g. 22]
  - Project Code Version

[AB#79271](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79271)

Eisie96:
## PR Type

[ x ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

If the 'quoting' feature is enabled, then all pages, where quoting components are rendered (/account, /account/quotes), are not stable. This problem is caused by the automatic refresh of quoting entities with the use of the 'timer' operator in the quotingEntities$ observable pipe.

Issue Number: Closes #1264

## What Is the New Behavior?

The quotingEntities$ function has a new automaticRefresh option. When the option is enabled, the timer operator will be executed right after the app is stable. 

[ ] Yes
[ x ] No

## Other Information

If the automaticRefresh option is enabled, it is necessary to subscribe to the quotingEntities$ observable in the component and to manually trigger the change detection. The reason for that is, that the used ApplicationRef.isStable observable runs outside of the Angular zone. It will not trigger the change detection automatically and not update the template.(https://docs.angular.lat/api/core/ApplicationRef)

Example from documentation:

```typescript
constructor(appRef: ApplicationRef, cd: ChangeDetectorRef) {
  appRef.isStable.pipe(
    first(stable => stable),
    switchMap(() => interval(1000))
  ).subscribe(counter => {
    this.value = counter;
    cd.detectChanges();
  });
}
```

[AB#79432](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79432)

### DIFF
--- a/src/app/extensions/quoting/facades/quoting.facade.ts
+++ b/src/app/extensions/quoting/facades/quoting.facade.ts
@@ -1,9 +1,9 @@
-import { Injectable } from '@angular/core';
+import { ApplicationRef, Injectable } from '@angular/core';
 import { Store, select } from '@ngrx/store';
-import { timer } from 'rxjs';
+import { Observable, OperatorFunction, identity, timer } from 'rxjs';
 import { map, sample, switchMap, tap } from 'rxjs/operators';
 
-import { whenFalsy } from 'ish-core/utils/operators';
+import { delayUntil, whenFalsy, whenTruthy } from 'ish-core/utils/operators';
 
 import { QuotingHelper } from '../models/quoting/quoting.helper';
 import { Quote, QuoteRequest, QuotingEntity } from '../models/quoting/quoting.model';
@@ -21,28 +21,18 @@ import {
 /* eslint-disable @typescript-eslint/member-ordering */
 @Injectable({ providedIn: 'root' })
 export class QuotingFacade {
-  constructor(private store: Store) {}
+  constructor(private store: Store, private appRef: ApplicationRef) {}
 
   loading$ = this.store.pipe(select(getQuotingLoading));
 
-  quotingEntities$() {
+  quotingEntities$(options?: { automaticRefresh: boolean }) {
     // update on subscription
     this.loadQuoting();
 
     return this.store.pipe(
       select(getQuotingEntities),
       sample(this.loading$.pipe(whenFalsy())),
-      switchMap(entities =>
-        // update every minute
-        timer(0, 60_000).pipe(
-          tap(count => {
-            if (count) {
-              this.loadQuoting();
-            }
-          }),
-          map(() => entities)
-        )
-      ),
+      options?.automaticRefresh ? this.automaticQuoteRefresh() : identity,
       tap(entities => {
         entities.filter(QuotingHelper.isStub).forEach(entity => {
           this.store.dispatch(loadQuotingDetail({ entity, level: 'List' }));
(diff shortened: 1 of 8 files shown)


src/app/extensions/quoting/pages/quote-list/quote-list-page.component.ts
```diff
@@ -9,14 +9,27 @@ import { Quote, QuoteRequest } from '../../models/quoting/quoting.model';
...
+  constructor(private quotingFacade: QuotingFacade, private cd: ChangeDetectorRef) {}
 
   ngOnInit() {
-    this.quotes$ = this.quotingFacade.quotingEntities$();
+    this.quotingFacade
+      .quotingEntities$({ automaticRefresh: true })
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(quotes => {
+        this.quotes = quotes;
+        this.cd.detectChanges();
```
SGrueber, line 26: try to use ngrxPush pipeline instead ( https://v11.ngrx.io/guide/component/push )

SGrueber (changes requested):

src/app/extensions/quoting/models/quoting/quoting.interface.ts
```diff
@@ -55,3 +55,7 @@ export interface QuoteRequestResponse extends QuoteBaseData<Link | QuoteRequestL
 }
 
 export type QuoteData = Link | QuoteResponse | QuoteRequestResponse;
+
+export interface QuoteEntitiesOptions {
```
SGrueber, line 59: Interfaces should only be used to accept data of REST calls. Anyway, I 'd prefer to remove it at all.
If use use an interface only at one place you can define it at this place, otherwise in the model.

src/app/extensions/quoting/facades/quoting.facade.ts
```diff
@@ -28,7 +29,7 @@ export class QuotingFacade {
   }
   loading$ = this.store.pipe(select(getQuotingLoading));
 
-  quotingEntities$(options?: { automaticRefresh: boolean }) {
+  quotingEntities$(options: QuoteEntitiesOptions = { automaticRefresh: true }) {
```
SGrueber, line 32: I would not use an options object here if it contains only one attribute. 

quotingEntities$( automaticRefresh: true ) ...

...would be fine.

SGrueber (changes requested):

src/app/extensions/quoting/facades/quoting.facade.ts
Eisie96: That was my first solution. I thought that the quotingEntities$ method with an object as optional parameter would be the better solution. The expression of ```this.facade.quotingEntities$({ automaticRefresh: false })``` is more explicit than ```this.facade.quotingEntities$(false)``` and understandable, what the method should do.

SGrueber (approved):

PR #1268, opened 2022-09-09
refactor: reuse default product listing type for mobile

dhhyi:
## PR Type

#1243 introduced a new environment config variable to handle default listing view types for mobile, while the existing variable could have been extended,

## What Is the New Behavior?

- Existing variable is extended.
- Existing variable is made optional.

[ ] Yes
[x] No

## Other Information

[AB#79484](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79484)


shauke (changes requested):
A documented example for how to set the view type for the two different device types would be nice in the `environment.model.ts` for those of us that cannot read the matrix. ;)

shauke (approved):
👍 Very nice with this documentation. 😄

PR #1271, opened 2022-09-10
fix(eslint-rules): repair ban-imports-file-pattern rule

dhhyi:
## PR Type

[x] Bugfix

In the process of migrating to eslint (#836), the rule ban-imports-file-pattern rule lost its ability to support regex patterns in the `name` field, which left some of the migrated configuration without effect.

## What Is the New Behavior?

- `name` supports patterns again
- applied rules or changed configuration

[ ] Yes
[ ] No

## Other Information

[AB#79500](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79500)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -241,9 +241,9 @@
               "message": "Modules should only aggregate deeper lying artifacts."
             },
             {
-              "name": "^(?!.*\\.(module|guard)$)\\.\\..*$",
+              "name": "^(?!.*\\.(module|guard|service)$)\\.\\..*$",
               "filePattern": ".*-routing\\.module\\.ts",
-              "message": "Modules should only aggregate deeper lying artifacts."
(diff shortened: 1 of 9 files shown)


SGrueber:
@dhhyi Thx for your contributions. We haven't noticed the issue related to the ban import files so far. Besides this we decided to revert your changes concerning the addToBasketItemSuccess action, because we want to keep the error parameter optional. It might be confusing if a success action starts with the httpError method and the error parameter is really only used in case adding one of the items failed.

dhhyi:
@SGrueber 
> ... because we want to keep the error parameter optional. It might be confusing if a success action starts with the httpError method and the error parameter is really only used in case adding one of the items failed.

Agreed, but then it should be better to just transfer the causes and not the complete error.

SGrueber (approved):

PR #1273, opened 2022-09-14
fix: repair dead-code script exclude comments search

### LINKED ISSUE #1222: find-dead-code script exception do not work for variable declarations
**Actual Behavior**

Running the find-dead-code script would fail for the following two variable definitions even though they are supposed to be exempt from the scripts testing.

```
// not-dead-code
export const reservedCharactersRegEx = /[ &\(\)=]/g;
```
https://github.com/intershop/intershop-pwa/blob/develop/src/app/core/utils/routing.ts#L31-L32

```
// not-dead-code
export const storeDevtoolsModule: unknown[] = [];
```
https://github.com/intershop/intershop-pwa/blob/develop/src/app/core/store/store-devtools.module.production.ts#L1-L2

We currently have a workaround in place that exempts the whole files from checking for dead code. This should not be necessary.
https://github.com/intershop/intershop-pwa/blob/develop/scripts/find-dead-code.ts#L7

**Expected Behavior**

Running the find-dead-code script should result in no errors with this configuration:
```
const fileExceptionsRegex = /\/src\/environments\//;
```

* Change the `fileExceptionsRegex` in `\scripts\find-dead-code.ts` to 
```
const fileExceptionsRegex = /\/src\/environments\//;
```
* Run
```
npx ts-node scripts/find-dead-code.ts
```

For some reason the logic of
```
  const ignoreComment = node.getPreviousSiblingIfKind(SyntaxKind.SingleLineCommentTrivia);
  if (ignoreComment?.getText().includes('not-dead-code')) {
    return;
  }
```
does not apply to the two mentioned `const` declarations.

[AB#78338](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/78338)

dhhyi:
## PR Type

[x] Bugfix

Issue Number: Closes #1222

## What Is the New Behavior?

- special case for finding exclude comment on VariableStatements
- cleanup unused dead-code excludes

[ ] Yes
[x] No

## Other Information

[AB#79590](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/79590)

### DIFF
--- a/schematics/src/extension/files/__name@dasherize__/store/__name@dasherize__-store.module.ts.template
+++ b/schematics/src/extension/files/__name@dasherize__/store/__name@dasherize__-store.module.ts.template
@@ -9,7 +9,6 @@ const <%= camelize(name) %>Reducers: ActionReducerMap<<%= classify(name) %>State
 
 const <%= camelize(name) %>Effects = [];
 
-// not-dead-code
 @NgModule({
   imports: [EffectsModule.forFeature(<%= camelize(name) %>Effects), StoreModule.forFeature('<%= camelize(name) %>', <%= camelize(name) %>Reducers)],
 })
--- a/scripts/find-dead-code.ts
+++ b/scripts/find-dead-code.ts
@@ -3,8 +3,8 @@ import { ClassDeclaration, Node, Project, ReferenceFindableNode, SyntaxKind, ts 
 
 /* eslint-disable no-console */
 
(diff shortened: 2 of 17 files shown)


SGrueber (approved):

shauke (approved):
Thanks for the script fix. 👍

Issue #625, opened 2021-03-19
Theme-dependent not found images


dhhyi:
**Is your feature request related to a problem? If yes, please describe it.**

Currently the image-not-found image (https://github.com/intershop/intershop-pwa/blob/develop/src/assets/img/not_available.png) is hard-coded in `ish-product-image`:
https://github.com/intershop/intershop-pwa/blob/dca53858266cac33ff956aa52562018d66db5be1/src/app/shared/components/product/product-image/product-image.component.html#L15

This makes it hard to set the image depending on the theme with the PWA's multi-theme support.

**Describe the desired solution.**

- Image source should be set in theme SCSS.
- It should be possible to supply SVG images and have them automatically scale to their surrounding containers.
- **BONUS** The text "NO IMAGE AVAILABLE" should be managed in a translation key and painted on top of the image when rendering, so it can be displayed in the current language.

**Describe alternatives you've considered.**

**Provide additional context.**

Additional images that are statically referenced:

![image](https://user-images.githubusercontent.com/23452927/111875083-34467880-8998-11eb-93ae-d02ab2f063fe.png)

andreassteinmann:
It would also be possible to replace the image using pure CSS styling (a circle with a crossing line) and to add the localized text below it. So you have no image at all and there would be no problem with any theme based image reference.

PR #1296, opened 2022-10-12
fix: replace the PNG image with a self explaining SVG image without text

andreassteinmann:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently, the "Image Not Available" image includes text within the image. This should be prevented due to localization. 

Issue Number: Relates to #625

## What Is the New Behavior?

- The PNG image is replaced by an SVG image. 
- The SVG image does not include text within the image any longer.
- The SVG image is self explaining and there is no need to add localized text. Furthermore, if the text would be added using pure HTML text it is hardly readable on thumbnail images (minicart, thumbs in product detail page etc.).
- The SVG image replaces the PNG in the same folder. It is not moved to a theme specific folder and the image reference is kept in the HTML component file(s) instead of SCSS. 

[x] Yes
[ ] No

## Other Information

The "Product Image Not Available" image is used for 
- product images 
- but also as a placeholder image for category images (until the category image is loaded)

Therefore both occurrences have to be tested.

### How to test
Create a new product without product images and test the display of the "product image not available" SVG on the following pages: 

- Product List
- PDP
- Minicart
- Cart
- Checkout Review
- Checkout Confirmation
- My Account - Order / Requisition Details
- My Account - Approval Details
- My Account - Order Template

[AB#80263](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/80263)


shauke (changes requested):
Everything looks fine now, except for the Firefox rendering issue.

image

shauke (approved):

PR #1313, opened 2022-10-28
docs: extend PurgeCSS integration documentation

shauke:
## PR Type

[x] Documentation content changes - regarding the integration and usage of PurgeCSS

>[PurgeCSS](https://purgecss.com/) is used to remove unused CSS classes from the CSS output. [Configuration](https://purgecss.com/configuration.html), especially [safelisting](https://purgecss.com/safelisting.html) certain classes, can be done on the plugin configuration or directly in your CSS with a special comment.

Seems to be not enough as documentation. 😉 

[x] No

## Other Information

[AB#80706](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/80706)

### DIFF
--- a/docs/guides/optimizations.md
+++ b/docs/guides/optimizations.md
@@ -50,8 +50,8 @@ The PurgeCSS plugin configuration can be found in the project's [`webpack.custom
 This method is used and recommended to include required styles of the third-party libraries used, which would otherwise be purged.
 For the different [configuration options](https://purgecss.com/configuration.html), refer to the PurgeCSS documentation.
 
-To protect styles defined in the Intershop PWA project source code, Intershop recommends safelisting them directly in your SCSS/CSS with [special comments](https://purgecss.com/safelisting.html#in-the-css-directly). 
-To include nested SCSS definitions, use `/* purgecss start ignore */` and `/* purgecss end ignore */` .
+To protect styles defined in the Intershop PWA project source code, Intershop recommends safelisting them directly in your SCSS/CSS with [special comments](https://purgecss.com/safelisting.html#in-the-css-directly).
+To include nested SCSS definitions, use `/* purgecss start ignore */` and `/* purgecss end ignore */`.
 
 ### Development
 
--- a/docs/guides/optimizations.md
+++ b/docs/guides/optimizations.md
@@ -51,7 +51,7 @@ This method is used and recommended to include required styles of the third-part
 For the different [configuration options](https://purgecss.com/configuration.html), refer to the PurgeCSS documentation.
 
 To protect styles defined in the Intershop PWA project source code, Intershop recommends safelisting them directly in your SCSS/CSS with [special comments](https://purgecss.com/safelisting.html#in-the-css-directly). 
-To include nested SCSS definitions using `/* purgecss start ignore */` and `/* purgecss end ignore */` seems to be the best way.
+To include nested SCSS definitions, use `/* purgecss start ignore */` and `/* purgecss end ignore */` .
 
 ### Development
 
--- a/docs/guides/optimizations.md
+++ b/docs/guides/optimizations.md
@@ -34,38 +34,36 @@ If the PWA is built using `production` configuration. (Either by building with `
 
 > This is where PurgeCSS comes into play. PurgeCSS analyzes your content and your CSS files. Then it matches the selectors used in your files with the one in your content files. It removes unused selectors from your CSS, resulting in smaller CSS files.
 
-While the described feature of purging unused CSS styles is wanted the mechanism of determining used styles is not completely without problems.
+While the described function for deleting unused CSS styles is very helpful, the mechanism for determining which styles are used is not without problems.
 PurgeCSS can only analyze the strings in the actual source code of the project for used styles.
-So styles that get added to the rendered HTML by third-party libraries (e.g.
-Bootstrap, Swiper) would not be found.
-The same goes for styles used in content loaded from the server (e.g.
-CMS, product descriptions).
-Also style selectors that are generated dynamically would not be found.
+
+So, styles that get added to the rendered HTML by third-party libraries (e.g. Bootstrap, Swiper) would not be found.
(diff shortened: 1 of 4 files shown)


docs/guides/optimizations.md
```diff
@@ -27,7 +27,45 @@ If the PWA is built using `production` configuration. (Either by building with `
...
 - All `data-testing` attributes are removed from the HTML templates to reduce output.
 - [PurgeCSS](https://purgecss.com) is used to remove unused CSS classes from the CSS output.
-  [Configuration](https://purgecss.com/configuration.html), especially [safelisting](https://purgecss.com/safelisting.html) certain classes, can be done on the plugin configuration or directly in your CSS with a special comment.
+
+## PurgeCSS
+
+> PurgeCSS is a tool to remove unused CSS. It can be part of your development workflow. When you are building a website, you might decide to use a CSS framework like TailwindCSS, Bootstrap, MaterializeCSS, Foundation, etc... But you will only use a small set of the framework, and a lot of unused CSS styles will be included.
+
+> This is where PurgeCSS comes into play. PurgeCSS analyzes your content and your CSS files. Then it matches the selectors used in your files with the one in your content files. It removes unused selectors from your CSS, resulting in smaller CSS files.
```
SGrueber, line 35: with the ones ?

docs/guides/optimizations.md
```diff
@@ -27,7 +27,45 @@ If the PWA is built using `production` configuration. (Either by building with `
...
+The same goes for styles used in content loaded from the server (e.g.
+CMS, product descriptions).
+Also style selectors that are generated dynamically would not be found.
+
+### Safelisting
+
+To solve this problem PurgeCSS provides different [options for safelisting](https://purgecss.com/safelisting.html) specific styles.
+This can either be done in the plugin configuration or directly in your SCSS/CSS files with special comments.
+
+The PurgeCSS plugin configuration can be found in the projects [`webpack.custom.ts`](https://github.com/intershop/intershop-pwa/blob/3.1.0/templates/webpack/webpack.custom.ts#L231-L246).
```
SGrueber, line 50: Maybe is it better to refer to a section/function in the file than to certain lines of code, because if they change we won't consider to change this documentation.

docs/guides/optimizations.md
```diff
@@ -27,7 +27,45 @@ If the PWA is built using `production` configuration. (Either by building with `
...
+This way is used and recommended to include needed styles of used third-party libraries that would otherwise be purged.
+The different [configuration options](https://purgecss.com/configuration.html) are described in the PurgeCSS documentation.
+
+To protect styles defined in the Intershop PWA project source code safelisting them directly in your SCSS/CSS with [special comments](https://purgecss.com/safelisting.html#in-the-css-directly) is recommended.
+To included nested SCSS definitions using `/* purgecss start ignore */` and `/* purgecss end ignore */` seems to be the best way.
```
SGrueber, line 55: To include ?

docs/guides/optimizations.md
```diff
@@ -27,7 +27,45 @@ If the PWA is built using `production` configuration. (Either by building with `
...
+
+### Development
+
+When using the standard way of developing the PWA with `ng s` PurgeCSS is not activated and styling should work as expected.
```
SGrueber, line 59: Nice documentation, next time a consultant/partner experiences problems I will refer to it :-)

SGrueber (changes requested):

docs/guides/optimizations.md
marschmidt89: This is a quote from their homepage apparently, I wouldn't change it. See https://purgecss.com/introduction.html

docs/guides/optimizations.md
```diff
@@ -27,7 +27,43 @@ If the PWA is built using `production` configuration. (Either by building with `
...
+
+To solve this problem PurgeCSS provides different [options for safelisting](https://purgecss.com/safelisting.html) specific styles.
+This can either be done in the plugin configuration or directly in your SCSS/CSS files with special comments.
+
+The PurgeCSS plugin configuration can be found in the project's [`webpack.custom.ts`](https://github.com/intershop/intershop-pwa/blob/3.1.0/templates/webpack/webpack.custom.ts#L231-L246).
+This method is used and recommended to include required styles of the third-party libraries used, which would otherwise be purged.
+For the different [configuration options](https://purgecss.com/configuration.html), refer to the PurgeCSS documentation.
+
+To protect styles defined in the Intershop PWA project source code, Intershop recommends safelisting them directly in your SCSS/CSS with [special comments](https://purgecss.com/safelisting.html#in-the-css-directly). 
+To include nested SCSS definitions using `/* purgecss start ignore */` and `/* purgecss end ignore */` seems to be the best way.
```
marschmidt89, line 54:
```suggestion
To include nested SCSS definitions, use `/* purgecss start ignore */` and `/* purgecss end ignore */` .
```
Not really happy with that "seems to be the best way" - sounds like we are not sure or haven't tested it enough - what do you think @shauke ?

marschmidt89 (dismissed):

docs/guides/optimizations.md
shauke: Unfortunately I know no way to reference a certain function in a source code file, that's why I choose to use specific lines.
To circumvent the problem with changing code lines and not longer fitting references I referenced a fixed version of the webpack.custom.ts `.../intershop-pwa/blob/3.1.0/...`. So we do not actually need to adapt the documentation except we (re)move the functionality.

docs/guides/optimizations.md
shauke: Well, I have tested it and found out that this is the safest way. So yes we can use your changed sentence.
The reason for phrasing it this way was to not being to bossy in regards to the different comment ways PurgeCSS provides. I wanted the developers to be more open to the original PurgeCSS documentation and experiment with it themselves. But yes this is not necessary. ;)

PR #1325, opened 2022-11-17
feat: save desired delivery date at basket items

SGrueber:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The desired delivery date is only saved at the basket as custom attribute 'desiredDeliveryDate'. In a standard ICM back office it is not shown.

Issue Number: Closes #

## What Is the New Behavior?
After the user has entered a desired delivery date on checkout shipping page it is additionally saved at all basket items. This way it is shown in the ICM back office in the order items table and can also be transferred to other back end systems like the IOM.

**Limitation**: In case of large baskets (> 20 items) saving the desired delivery date at all basket items might cause (unacceptable) long response times, a possible solution for this would be to implement this feature in the  ICM back end. Maybe in future there will also be a bulk REST call to overcome this.

[ ] Yes
[x] No

## Other Information
This feature requires an ICM 7.10.39 or higher

[AB#81176](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/81176)

### DIFF
--- a/src/app/core/store/customer/basket/basket.effects.ts
+++ b/src/app/core/store/customer/basket/basket.effects.ts
@@ -16,6 +16,7 @@ import {
   withLatestFrom,
 } from 'rxjs/operators';
 
+import { AttributeHelper } from 'ish-core/models/attribute/attribute.helper';
 import { Basket } from 'ish-core/models/basket/basket.model';
 import { BasketService } from 'ish-core/services/basket/basket.service';
 import { getCurrentCurrency } from 'ish-core/store/core/configuration';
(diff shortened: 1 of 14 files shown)


src/app/core/store/customer/basket/basket.effects.ts
```diff
@@ -197,6 +203,30 @@ export class BasketEffects {
...
   );
 
+  /**
+   * Sets a desired delivery date at the current basket and each line item.
+   */
+  setBasketDesiredDeliveryDate$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(setBasketDesiredDeliveryDate),
+      mapToPayloadProperty('desiredDeliveryDate'),
+      concatMap(date =>
```
Eisie96, line 213: Does the rxjs operator concatMap make sense for setting a basket delivery date? The stream would wait for setting a new desiredDeliveryDate property until the following basket service observable completed.

Eisie96 (changes requested):

PR #1328, opened 2022-11-22
refactor: add enum CheckoutTargetStep to set next checkout step

Eisie96:
## PR Type

[ ] Bugfix
[ ] Feature
[ x ] Refactoring (no functional changes, no API changes)
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The checkout target step is set with a number. It is not really readable and understandable.

Issue Number: Closes #

## What Is the New Behavior?

The enum 'CheckoutStepType' is implemented, which represents the available, ordered checkout steps. It should improve the understanding of the code.

[ ] Yes
[ x ] No

## Other Information

[AB#81268](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/81268)

### DIFF
--- a/src/app/core/models/checkout/checkout-step.type.ts
+++ b/src/app/core/models/checkout/checkout-step.type.ts
@@ -2,9 +2,9 @@
  * available, ordered checkout steps
  */
 export enum CheckoutStepType {
-  addresses = 1,
-  shipping,
-  payment,
-  review,
-  receipt,
+  Addresses = 1,
+  Shipping,
+  Payment,
+  Review,
+  Receipt,
 }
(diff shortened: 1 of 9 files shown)


src/app/core/models/checkout/checkout-step.type.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * available, ordered checkout steps
+ */
+export enum CheckoutStepType {
+  addresses = 1,
```
SGrueber, line 5: Because these are like constants I would prefer to write them with a starting capital letter as it is in all other occurances, see e.g. here: https://www.typescriptlang.org/docs/handbook/enums.html

SGrueber (changes requested):
This is a nice rework, it is much more readable now !

PR #1337, opened 2022-12-08
feat: extension address doctor

carlosDjangoo:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

There is no Address Doctor Extension to check if the entered address is valid.

Issue Number: Closes #

## What Is the New Behavior?

This feature is intended to integrate the Address Doctor address validation web service. If the feature is enabled, a call to the web service will be executed whenever a new address needs to be validated.  After clicking the save button an Modal overlay will appear and you can choose the address you want. You can choose your entered or a suggestion from Address Doctor.

[ ] Yes
[x] No

## Other Information

address-doctor

[AB#81763](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/81763)


marschmidt89 (commented):
Localization changes approved

mglatter (approved):
Localizations approved

src/app/extensions/address-doctor/models/address-doctor-config.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface AddressDoctorConfig {
+  login: string;
+  password: string;
```
charleshopfner, line 3: Hi,
I can't really find the recommendations from Address Doctor to use their apis, is it secure to store in the front the login and password to use their services ? Isn't there a risk for someone to collect it ?

shauke: @charleshopfner The client needs to send the login and password with the POST REST request in plain text to the Address Doctor server. So the client will have to know these values at some point.
In a production environment these values are set via an environment variable at deployment time, so the values are not part of the actual source code and we assume that is "secret" enough.

But we are open to any suggestions how this could be handled better if necessary.

shauke (approved):

shauke (approved):

PR #1372, opened 2023-02-16
fix: validation for passwort and email confirmation fields

Eisie96:
## PR Type

[ x ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

For all confirmation fields the `ish-password-field` or `ish-email-field` is used. These fields are configured with a validator, which checks, that the form value follows some specified rules. This is not intended, because a confirmation input field should only check, if the value equals the form value of a related field.

## What Is the New Behavior?

All confirmation fields are configured with the default `ish-text-input` field type. All necessary validators are added seperatly. Furthermore the missing tranlation key `account.update_email.email.error.notempty` is added.

[ ] Yes
[ x ] No

## Other Information

[AB#83799](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/83799)

### DIFF
--- a/src/app/extensions/punchout/shared/punchout-user-form/punchout-user-form.component.ts
+++ b/src/app/extensions/punchout/shared/punchout-user-form/punchout-user-form.component.ts
@@ -92,8 +92,9 @@ export class PunchoutUserFormComponent implements OnInit {
           },
           {
             key: 'passwordConfirmation',
-            type: 'ish-password-field',
+            type: 'ish-text-input-field',
             props: {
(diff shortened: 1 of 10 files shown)


Eisie96:
This is not needed, because the translation keys for a default email confirmation error message is not used. The keys will be removed in the next commit.

SGrueber (dismissed):

SGrueber (approved):

PR #1376, opened 2023-02-23
feat: print styling

DilaraGueler:
## PR Type

[X] Feature

No print rules exist.

## What Is the New Behavior?

General print rules were introduced to the application without a header, and footer, with a reasonable layout, and nice page breaks.

[X] No

## Other Information

[AB#83954](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/83954)

### DIFF
--- a/projects/requisition-management/src/app/pages/requisition-detail/requisition-detail-page.component.html
+++ b/projects/requisition-management/src/app/pages/requisition-detail/requisition-detail-page.component.html
@@ -115,13 +115,13 @@
     ></ish-requisition-reject-dialog>
 
     <div class="section d-flex d-flex justify-content-between">
-      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]" class="noprint">
+      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]" class="no-print">
         <ng-container *ngIf="view === 'buyer'; else backToApprovals">
           {{ 'approval.detailspage.links.return_to_requisitions' | translate }}
         </ng-container>
         <ng-template #backToApprovals>{{ 'approval.detailspage.links.return_to_approvals' | translate }}</ng-template>
       </a>
-      <a routerLink="/home" class="noprint" data-testing-id="home-link">{{
+      <a routerLink="/home" class="no-print" data-testing-id="home-link">{{
         'account.orderdetails.links.continue_shopping' | translate
       }}</a>
     </div>
--- a/projects/requisition-management/src/app/pages/requisition-detail/requisition-detail-page.component.html
+++ b/projects/requisition-management/src/app/pages/requisition-detail/requisition-detail-page.component.html
@@ -115,13 +115,13 @@
     ></ish-requisition-reject-dialog>
 
     <div class="section d-flex d-flex justify-content-between">
-      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]">
+      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]" class="noprint">
         <ng-container *ngIf="view === 'buyer'; else backToApprovals">
           {{ 'approval.detailspage.links.return_to_requisitions' | translate }}
         </ng-container>
         <ng-template #backToApprovals>{{ 'approval.detailspage.links.return_to_approvals' | translate }}</ng-template>
       </a>
-      <a routerLink="/home" data-testing-id="home-link">{{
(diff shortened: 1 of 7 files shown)


projects/requisition-management/src/app/pages/requisition-detail/requisition-detail-page.component.html
```diff
@@ -115,13 +115,13 @@ <h3>{{ 'checkout.order_summary.heading' | translate }}</h3>
     ></ish-requisition-reject-dialog>
 
     <div class="section d-flex d-flex justify-content-between">
-      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]">
+      <a [routerLink]="['../../' + view, { status: requisition.approval.statusCode }]" class="noprint">
```
andreassteinmann, line 118: Please rename the class to `no-print` (and change at all places in HTML an SCSS.

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
...
 // CUSTOMIZED PRINT STYLES
+@import 'variables';
 
 @media print {
+  @page {
+    margin-top: 100px;
+  }
+
+  @supports (-moz-appearance: none) {
+    .logo {
```
andreassteinmann, line 11: Please put the logo styling at one place (not at the top and the bottom of the file).

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
 //
 // CUSTOMIZED PRINT STYLES
+@import 'variables';
 
 @media print {
+  @page {
+    margin-top: 100px;
```
andreassteinmann, line 7: Please consider adding a paper size like A4 that the user is able to select portrait and landscape mode.

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
 //
 // CUSTOMIZED PRINT STYLES
```
andreassteinmann, line 2: Please add a general description like "apply general print styles ... for PDF creation ... apply to pages with a print icon"

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
...
+    .list-item-row {
+      margin-right: 2.5%;
+    }
+  }
+
+  .line-item-list,
+  .order-summary {
+    display: inline-block;
+    break-before: page !important;
+    page-break-before: always !important; /* Older browser versions */
```
andreassteinmann, line 88: change "Older" to "older" (lowercase)

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
...
+    .logo {
+      margin-left: 25px !important;
+    }
+  }
+
   a[href]::after {
     content: none !important;
   }
+
+  button {
```
andreassteinmann, line 20: Please move the button styling down to where the other button styling is applied. You should first have all items listed which are not displayed and afterwards the elements where a certain visual styling is applied. It makes it easier to read the CSS file.

src/styles/global/print.scss
```diff
@@ -1,8 +1,96 @@
...
+  .fa-print,
+  .fa-trash-can,
+  .header-utility,
+  .marketing-area,
+  .navbar-toggler,
+  .product-tile-actions,
+  .share-label,
+  .toast-container,
+  .user-links {
+    display: none !important;
```
andreassteinmann, line 55: Good order of elements!

src/styles/global/print.scss
DilaraGueler: Tried different page size options. It breaks the whole design.

src/styles/global/print.scss
DilaraGueler: Thanks!

DilaraGueler:
> Is there any solution, to make the borders around the green fields not red but green? Because in this case the buyer is still in the green.
> 
> image

Done! The options were green, orange, and red. The borders will be colored according to the background colors.

DilaraGueler:
> The print document name is a little bit misleading. My Account_Intershop-PWA.pdf

Should be asked the PO, if file naming will be in the scope of this task.

src/styles/global/print.scss
andreassteinmann: True, I tried several approaches with the @page size but they always create unwanted layouts even in Chrome.

andreassteinmann (approved):

iwiederhold (approved):
I like it! THX

PR #1420, opened 2023-04-20
fix: handle case where all recently viewed products are failed/offline

janjoosse:
## PR Type

[x] Bugfix

When all recently viewed products that displayed in the `Recently viewed` section go offline, the PWA still renders this heading for `Recently viewed` even though there are no products displayed. 

Issue Number: Closes #

## What Is the New Behavior?
The `recently.facade.ts` now filters both the `recentlyViewedProducts` and `mostRecentlyViewedProducts`, excluding products that are in the failed products store, instead of doing the filtering inside the `products-list.component.ts`. 

[ ] Yes
[x] No

## Other Information

Issue described in support ticket https://support.intershop.com/cse/index.php/Display/967271

### DIFF
--- a/src/app/core/facades/shopping.facade.ts
+++ b/src/app/core/facades/shopping.facade.ts
@@ -28,7 +28,6 @@ import {
 import { loadProductPrices } from 'ish-core/store/shopping/product-prices';
 import { getProductPrice } from 'ish-core/store/shopping/product-prices/product-prices.selectors';
 import {
-  getFailedProducts,
   getProduct,
   getProductLinks,
   getProductParts,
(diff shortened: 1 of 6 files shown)


shauke:
@janjoosse Thank you for this contribution. It is definitely an improvement for the recently viewed product list case.
But we are using the `products-list.component` for other places as well were we rely on the functionality that it handles offline/failed products. So we will have to see if or how we get these two things together or if we should handle the check outside the `products-list.component` similar as you suggest for the recently viewed case.

Out of curiosity, what shop looses all its recently viewed products? The check was supposed to be more for handling a few exceptions. 😉

shauke:
The mentioned idea with the `shopping.facade.excludeFailedProducts$` was used and implemented on top of the general failed products filtering within the products list. The PR #1647 is merged instead of this PR.

PR #1421, opened 2023-04-21
feat: token behavior improvements

### LINKED ISSUE #1374: SSO becames a mess since 3.2.0
Related commit : https://github.com/intershop/intershop-pwa/commit/4e02efd4c8a12cd325d8ed853d4f8c258b8d9445
Since 3.2.0 intershop use  ICM '/token' REST endpoint for authentication and it relies heavily on angular-oauth2-oidc to handle lifecycle of ICM token.
Even for anonymous browsing an oauth token is required.

The problem starts when you try to integrate a third-party SSO system.
Now you have to handle a token from 2 sources, one for the anonymous browsing (ICM) and one for logged browser (external SSO system).

Kind of hack has been code there : 
https://github.com/intershop/intershop-pwa/blob/1f876847d4a41a4a219f86f756fb0e8d8f939977/src/app/core/identity-provider/auth0.identity-provider.ts#L90

But this configuration will be override when openid document is loaded. And that created a lots of side effets

**Expected Behavior**

Anonymous token should not be driven by SSO mecanism

[AB#83902](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/83902)

Eisie96:
## PR Type

[ ] Bugfix
[ x ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The new token api is introduced with the PWA version 3.2 (#1156). Since then, we are dealing with some issues, which needed to be addressed.

* The ICM token functionality is split into multiple files. The configuration of the OAuth service (identity-provider) is executed in a different place then fetching the token (user.service.ts). This can lead to a runtime issue, when the oAuthService wants to fetch a new token, although the whole service is not yet configured.
* The PWA can now be initialized with two different oAuthService combination. (Auth0 and default ICM token behavior). Right now all configurations are sharing just one oAuthService instance. This can lead to unneccary side effects (overriden configuration)
* The anonymous user token is fetched everytime no apiToken is available. This is not necessary. The token is only needed, when a new basket should be created.

Issue Number: Closes #1374

## What Is the New Behavior?

The `OauthConfigurationService` is replaced by the `TokenService`. The service will inject it's own OAuthService instance and is responsible for it's configuration and fetching token mechanism. All identity provider are not relying on the internal ICM token behavior anymore.

The anonymous user token behavior is adapted. It will only be fetched, when an anoymous user will create a basket. The functionality is included in the `ApiTokenService` intercept method.

The identity provider needs to be adapted. These providers are not responsible anymore to configure the oAuthService for fetching ICM tokens. 

When a identity provider needs a OAuthService (e.g. auth0-identity-provider), then it has to inject the dependency with a complete new OAuthService instance. Otherwise it could lead to unnecassary side effects. (e.g. during configuration)

```typescript
  export class Auth0IdentityProvider implements IdentityProvider {
    private oauthService: OAuthService;
  
    constructor(
      parent: Injector,
    ) {
      const injector = Injector.create({
        providers: [{ provide: OAuthService }],
        parent,
      });
  
      this.oauthService = injector.get(OAuthService);
    }
}
```

[ x ] Yes
[ ] No

## Other Information

[AB#85534](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/85534)


Eisie96:
@tbouliere-datasolution

Hello Tristan, could you have a look on this PR? We have improved the token handling in the PWA and need some further feedback from you. :)

tbouliere-datasolution:
Hi @Eisie96 
I haven't tested it thoroughly, but doesn't the following function  https://github.com/intershop/intershop-pwa/blob/11fd8155f595f67243b76ab1f22f98370e241842/src/app/core/services/token/token.service.ts#L92-L101  force an anonymous token ?

Won't this invalidate the sso token when this function is called?

shauke (changes requested):
Problem with anonymous apiToken cookie from ICM in Hybrid Approach

When switching back from an ICM rendered page in Hybrid Approach where an anonymous apiToken cookie was set without a basket, the PWA tries to fetch the basket for that cookie and fails (REST error).
This results an disabled "Add to Cart" buttons in the whole PWA. 

Other than this the rework looks good and worked with SSO and Punchout tests as expected.

shauke (dismissed):

shauke:
Fixed Hybrid Approach issue that was independent from the token behavior improvements. Had the same problems before.

shauke (approved):

PR #1425, opened 2023-04-28
fix: adapt 'addGlobalGuard' function to work with functional guards

shauke:
## PR Type

[x] Bugfix

The changes for "refactor(Angular 15): replace class-based route guards by functional guards" (https://github.com/intershop/intershop-pwa/commit/289e1616ce4093acb35b4e070291088693e1a192) broke the registration of the global `hybridRedirectGuard` needed for the Hybrid Approach. So the Hybrid Approach is not working with PWA 4.0.0.

## What Is the New Behavior?

The global guard registration works again and the Hybrid Approach can be used with current PWA versions again.

[x] No

## Other Information

[AB#85701](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/85701)

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -11,6 +11,9 @@ kb_sync_latest_only
 
 The Intershop PWA now uses Node.js 18.16.0 LTS with the corresponding npm version 9.5.1 to resolve an issue with Azure docker deployments (see #1416).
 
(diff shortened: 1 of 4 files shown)


SGrueber (approved):

SGrueber:
Are you sure the check for 'guard.prototype.canActivate' is not needed any more?

SGrueber (approved):

PR #1432, opened 2023-05-15
fix: use 'stepQauntitiy' instead of 'stepOrderQuantity' to fetch the correct data in product listings

DilaraGueler:
## PR Type

[x] Bugfix

The current ICM product detail REST API returns `stepOrderQuantity` information.
For that reason the product list calls where asked for `stepOrderQuantity` as well but did not return that information.
For product list calls only the better name `stepQuantity` works and returns the expected information.

## What Is the New Behavior?

The PWA now consistently uses the name `stepQuantity`.
Only for the product details call the mapper maps the `stepOrderQuantity` value to `stepQuantity`.
This way step quantity now works in product listings as well as on the product details page.

In the future the ICM product details REST API will provide the information as `stepQuantity` as well and the `stepOrderQuantity` will be deprecated. Once this is the case another adaption in the PWA will become necessary.

[x] No

## Other Information

[AB#86081](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/86081)


shauke (approved):
Adapted and approved

PR #1433, opened 2023-05-17
chore: emit 'NotVisible' events when target element moves out of viewport

janjoosse:
## PR Type

[X] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Currently the `ishIntersectionObserver` can be used as directive on a target element to determine whether the element is visible. Whenever the element moves into view and becomes visible it will emit the `IntersectionStatus` `Visible`. Given the type definition for `IntersectionStatus` [`export type IntersectionStatus = 'Visible' | 'Pending' | 'NotVisible';`] I would expect that the directive also emits a `NotVisible` event when the target element becomes invisible. This is not the case, it only emits `Pending` and `Visible` statuses.

## What Is the New Behavior?
The `ishIntersectionObserver` also emits `NotVisible` status.

[ ] Yes
[X] No

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -17,6 +17,9 @@ If the `addGlobalGuard` function is used for customizations make sure it now wor
 The input parameter `id` of the component `ish-product-quantity` caused issues with duplicate IDs within the page html, therefore it was renamed to `elementId`.
 If the input parameter 'id' of this component has already been used it has to be renamed accordingly.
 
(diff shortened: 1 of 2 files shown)


shauke (approved):
You are right, there is no need to filter the `NotVisible` events.
The usage in this project filters on the events themselves.

shauke (approved):

PR #1442, opened 2023-06-02
chore: add generic intersection types for ViewType & DeviceType

kepek:
It might be useful when building some custom view for other devices or when you're in need to create a different product listing view that requires a bit different look compared to the OOTB solution but still, you may want 
to remain everything else to be fully backward compatible.

```ts
export type MyViewType = ViewType<'simple' | 'detailed'>;

export type MyDeviceType = DeviceType<'wearable' | 'smartTv'>;
```

## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

Issue Number: Closes #

[ ] Yes
[x] No

### DIFF
--- a/src/app/core/models/viewtype/viewtype.types.ts
+++ b/src/app/core/models/viewtype/viewtype.types.ts
@@ -1,3 +1,3 @@
-export type ViewType = 'grid' | 'list';
+export type ViewType<T extends string = ''> = T | ('grid' | 'list');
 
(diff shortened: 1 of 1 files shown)


SGrueber:
@kepek Can you explain a bit more detailed why these changes are helpful? Would it not be easier to simply override the existing types and adapt them to your project needs?

PR #1461, opened 2023-07-12
fix: reload the application after the pgid changed

SGrueber:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

If the user's pgid changes due to data changes in the ICM backoffice, e.g. assigning/unassigning the user to a customer group, promotion etc. the personalized REST calls will fail with a 400 error: Bad Request (The provided matrix parameter "pgid" does not match the authenticated user.) and an error page is shown in most cases.

Issue Number: Closes #

## What Is the New Behavior?
There is a new error handler that checks for the error above and if this error occurs the current page is reloaded. The relaod is necessary because a lot of data in the ngrx store might be affected (invalid) and the reload forces to fetch all the data from the ICM server again.

[ ] Yes
[x] No

## Other Information

[AB#87744](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/87744)

### DIFF
--- a/src/app/core/core.module.ts
+++ b/src/app/core/core.module.ts
@@ -12,6 +12,7 @@ import { ICMErrorMapperInterceptor } from './interceptors/icm-error-mapper.inter
 import { IdentityProviderInterceptor } from './interceptors/identity-provider.interceptor';
 import { MockInterceptor } from './interceptors/mock.interceptor';
 import { PaymentPayoneInterceptor } from './interceptors/payment-payone.interceptor';
+import { PGIDChangeInterceptor } from './interceptors/pgid-change.interceptor';
 import { PreviewInterceptor } from './interceptors/preview.interceptor';
 import { InternationalizationModule } from './internationalization.module';
 import { StateManagementModule } from './state-management.module';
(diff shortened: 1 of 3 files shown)


src/app/core/utils/http-error/pgid-change.error-handler.ts
```diff
@@ -0,0 +1,11 @@
+import { SpecialHttpErrorHandler } from 'ish-core/interceptors/icm-error-mapper.interceptor';
```
Eisie96, line 1: I don't think that the SpecialHttpErrorHandler is the correct interceptor to react on icm call failure due to outdated pgid. It was introduced to map errors to specific error codes. I think that a custom `HttpInterceptor` is the better selection.

Eisie96 (changes requested):

src/app/core/utils/http-error/pgid-change.error-handler.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// reload page when the pgid of the user has changed
+export const pgidChangeErrorHandler: SpecialHttpErrorHandler = {
+  test: error =>
+    error.error === 'Bad Request (The provided matrix parameter "pgid" does not match the authenticated user.)',
+  map: () => {
+    window.location.reload();
```
Eisie96, line 8: If a HttpInterceptor should be introduced, than it makes sense to reload the page only on browser side.

Eisie96 (changes requested):

src/app/core/utils/http-error/pgid-change.error-handler.ts
SGrueber: you won't get this error for SSR requests

PR #1468, opened 2023-07-28
Update build-pwa.js - use defualtProject instead of hardcoded name

kepek:
It might be useful when you want to build a 100% custom build based on `intershop-pwa` which is also part of the local angular.json configuration under a different project name.

## PR Type

[ ] Bugfix
[ ] Feature
[x] Build-related changes
[ ] CI-related changes
[ ] Other: 

Issue Number: Closes #

[ ] Yes
[x] No

### DIFF
--- a/scripts/build-pwa.js
+++ b/scripts/build-pwa.js
@@ -2,6 +2,8 @@ const fs = require('fs');
 const path = require('path');
 const execSync = require('child_process').execSync;
 
+const angularJson = JSON.parse(fs.readFileSync('./angular.json', { encoding: 'utf-8' }));
+
 /**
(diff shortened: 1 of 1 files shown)


SGrueber:
Since angular cli 14 the defaultProject property in the angular.json has been deprecated and with angular 16 it has been removed.
This means, we have to find a replacement for it and will use a different mechanism to define the default project in the near future, most probably with the angular 16 update ( see #1467 ). We will consider your changes if we refactor the defaultProject option.

SGrueber:
We removed the default project from the angular.json in this [commit]('https://github.com/intershop/intershop-pwa/commit/83b9a37017cd998b0c2c9e7c93bbe8142d669653')

PR #1469, opened 2023-08-03
feat: shared nginx cache with memcached

dhhyi:
## PR Type

[x] Feature

In a classic PWA deployment, nginx instances each have their own cache embedded. If the deployment runs multiple nginx instances, the cache is not synchronized, so a request could hit a cache entry when asking server one but not server two.

## What Is the New Behavior?

The cache among nginx instances is shared using a memcached instance running alongside.

[x] Yes
[ ] No

Caching embedded is replaced by caching with memcached. Different cache config durations for 200 and 404 responses are no longer possible.

## Other Information

[AB#88303](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/88303)

## TODO

- [ ] repair CACHE_DURATION_NGINX_OK and CACHE_DURATION_NGINX_NF?
- [ ] make memcached opt-in?


philippsauer:
We have set up a test system prepared with 2 Nginx and related PWA adjustments: https://pwa-snnck-uatb.pwa.intershop.de/
In general, the fix seems to work, once cached pages are permanently delivered with X-Cache-Fetch-Status: HIT.

But there are problems with the styles, it permanently returns 304: Not Modified. See example storefront (requires Intershop VPN): https://pwa-snnck-uatb.pwa.intershop.de/bigorder/home. Related request: https://pwa-snnck-uatb.pwa.intershop.de/styles.2f6d7e311aa158dc.css
image
 
Also some js files return 304:
image

Do you have an idea what could be the reason? Unchecking "Disable cache" in the dev tools didn't help unfortunately.

dhhyi:
> Do you have an idea what could be the reason?

We discovered that the nginx was trying to cache content bigger than 1MB in size, which was denied by memcached.
This is now fixed by the following modifications:
- static assets are cached in each nginx instance
- memcached item size is raised to 10M
- nginx (srcache) store limit is set to 10M
- nginx client body size check for communication with memchecked is disabled

PR #1475, opened 2023-08-08
feat(ICM 11): co-browsing support via CEC

SGrueber:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

## What Is the New Behavior?
Co-browsing feature enables an administrative Intershop Contact Center user to be able to login in into the pwa as a user with no need to have the credentials of that user giving him/her (the admin) the ability to see how the site looks from a real user perspective.
A new route 'cobrowse' has been introduced that has to be called together with a parameter 'token' to support this feature.

[ ] Yes
[x] No

## Pending Issues
- [ ] ICM release 11.8 is needed or ICM 7.10 without SSO
- [ ] the co-browse identity provider should always be active when the user enters the cobrowse route together with a valid token as query parameter. Currently, this works only properly with a multisite configuration

## Other Information

[AB#88392](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/88392)

### DIFF
--- a/src/app/core/identity-provider/co-browse.identity-provider.ts
+++ b/src/app/core/identity-provider/co-browse.identity-provider.ts
@@ -1,0 +1,122 @@
+import { HttpEvent, HttpHandler, HttpRequest } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { ActivatedRouteSnapshot, Router } from '@angular/router';
+import { Store, select } from '@ngrx/store';
+import { Observable, noop, of, race, throwError } from 'rxjs';
+import { catchError, concatMap, delay, filter, first, map, switchMap, take, withLatestFrom } from 'rxjs/operators';
+
+import { AccountFacade } from 'ish-core/facades/account.facade';
+import { AppFacade } from 'ish-core/facades/app.facade';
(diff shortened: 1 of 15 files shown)


src/app/core/identity-provider/co-browse.identity-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+        if (!apiToken) {
+          this.accountFacade.fetchAnonymousToken();
+        }
+        if (type === 'user') {
+          this.accountFacade.logoutUser({ revokeApiToken: false });
+        }
+      });
+
+    // OAuth Service should be configured before apiToken information are restored and the refresh token mechanism is setup
+    this.apiTokenService.restore$(['user', 'order']).subscribe(noop);
```
kepek, line 48: We have implemented something similar for our client and we have observed that token is not encoded by ICM URL, so we need to reinsert '+'
``` accessToken = decodeURIComponent(accessToken?.replace(/\s/g, '+'));```

shauke (approved):

PR #1477, opened 2023-08-11
feat: fallback to loading when encountering proxy timeout

dhhyi:
## PR Type

[x] Feature

High loads on SSR (as well as improper configuration or technical debt) can exhaust all rendering processes and users get long loading times or timeouts.

## What Is the New Behavior?

If a timeout is encountered, the loading page will be sent to the user so that Angular can boot up on the client. This does not give the user a full page response, but he can use the webshop.

[ ] Yes
[x] No

## TODO/Discussion

- [ ] make timeout configurable
- [ ] reduce code duplication
- [ ] assure the loading page is always available and not unavailable to high load as well
- [ ] research if it is possible to cache the timeout-ed response after it is finally rendered (with built-in methods, it is discarded), or if the SSR rendering can be cancelled

## Other Information

[AB#88480](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/88480)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -67,7 +67,9 @@ services:
     environment:
       UPSTREAM_PWA: 'http://pwa:4200'
       NGINX_ENTRYPOINT_QUIET_LOGS: ANYVALUE
-      CACHE: 0
+      # CACHE: 0
+      CACHE_DURATION_NGINX_OK: 20s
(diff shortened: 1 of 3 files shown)


dhhyi:
replaced by #1511

PR #1485, opened 2023-08-18
feat: implement better handling for cookie 'SameSite' and 'secure' settings

shauke:
## PR Type

[x] Feature

## What Is the New Behavior?

feat: implement better handling for cookie 'SameSite' and 'secure' settings

* defaults to 'SameSite=Strict' and 'secure' if not explicitly set to other values
* sets all cookies with 'SameSite=None' if run within an iframe
* if run with 'http' (should only be in development environments) 'secure' is not set
* inspired by https://medium.com/trabe/cookies-and-iframes-f7cca58b3b9e

[x] Yes

The `cookies.service` has new defaults: `SameSite=Strict` and `secure` (see [Migrations / 4.1 to 4.2](https://github.com/intershop/intershop-pwa/blob/develop/docs/guides/migrations.md#41-to-42) for more details).

## Other Information

[AB#88753](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/88753)

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -7,6 +7,17 @@ kb_sync_latest_only
 
 # Migrations
 
+## 4.1 to 4.2
+
+A better handling for cookie `SameSite` and `secure` settings was implemented with new defaults to `SameSite=Strict` and `secure`.
+This can still be overridden when calling `cookies.services` `put` method with explicitly set values.
+Now the `secure` setting is always set to `true` if in `https` mode, you can prevent this by explicitly setting it to `false`.
+If the PWA is run with `http` (should only be in development environments) `secure` is not set.
+Additionally if the PWA is run in an iframe all cookies are set with `SameSite=None` (e.g. for punchout or Design View).
+Be aware that some browsers no longer accept cookies with `SameSite=None` without `secure`.
+Before by default no `SameSite` was set so browsers treated it as `SameSite=Lax`, this needs to be set explicitly now if it is really intended.
+For migrating check the calls of the `cookies.service` `put` method whether they need to be adapted.
+
 ## 4.0 to 4.1
 
 The Intershop PWA now uses Node.js 18.16.0 LTS with the corresponding npm version 9.5.1 to resolve an issue with Azure Docker deployments (see #1416).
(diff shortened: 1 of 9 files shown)


docs/guides/migrations.md
```diff
@@ -7,6 +7,15 @@ kb_sync_latest_only
 
 # Migrations
 
+## 4.1 to 4.2
+
+A better handling for cookie `SameSite` and `secure` settings was implemented with new defaults to `SameSite=Strict` and `secure`.
+This can still be overridden when calling `cookies.services` `put` method with explicitly set values.
+Additionally if the PWA is run in an iframe all cookies are set with `SameSite=None` and if run with `http` (should only be in development environments) `secure` is not set.
+For migrating the occurrence of `cookies.services` `put` can be check whether they need to be adapted.
```
SGrueber, line 15: For migrating check the calls of `cookies.services` `put` method whether they need to be adapted. (maybe this should be the last sentence of this block)

docs/guides/migrations.md
```diff
@@ -7,6 +7,15 @@ kb_sync_latest_only
...
+Before by default no `SameSite` was set so browsers treated it as `SameSite=Lax`, this needs to be set explicitly now if it is really intended.
+The `secure` setting was only set if explicitly set to `secure: true` and if in `https` mode otherwise it was set to `false`.
```
SGrueber, line 17: Now the `secure` setting is always set to true if in `https` mode, you can prevent this by explicitly set it to 'false'.

docs/guides/migrations.md
```diff
@@ -7,6 +7,15 @@ kb_sync_latest_only
 
 # Migrations
 
+## 4.1 to 4.2
+
+A better handling for cookie `SameSite` and `secure` settings was implemented with new defaults to `SameSite=Strict` and `secure`.
+This can still be overridden when calling `cookies.services` `put` method with explicitly set values.
+Additionally if the PWA is run in an iframe all cookies are set with `SameSite=None` and if run with `http` (should only be in development environments) `secure` is not set.
```
SGrueber, line 14: May here (or elsewhere) we should mention that running the pwa in an iframe with http is not working properly and will lead to cookie errors.

SGrueber (changes requested):

docs/guides/migrations.md
shauke: This we need to discuss.

SGrueber (approved):

PR #1509, opened 2023-09-14
feat: shared nginx cache with redis

dhhyi:
## PR Type

[X] Feature

Each nginx has its own cache so in a deployment with multiple nginx (for redundancy), the cache hit rate is significantly lower than it could be.

## What Is the New Behavior?

Nginx instances push the cache to a shared redis service and retrieve it from there.

[ ] Yes
[X] No

## Other Information

[AB#89494](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/89494)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -22,6 +22,7 @@
 *.properties
 *.crt
 *.key
+*.lua
 .*ignore
 .husky
 .gitattributes
--- a/.vscode/intershop.txt
+++ b/.vscode/intershop.txt
@@ -11,6 +11,7 @@ appserver
 backoffice
 bar
 bars
+bitnami
 blocklist
 breakline
 categoryref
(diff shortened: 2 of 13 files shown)


philippsauer:
Hi, regarding the cache invalidation: what would be the best way to trigger that (periodically, manually) in a production environment without connecting via shell?

```bash
docker run --rm -it bitnami/redis redis-cli -u <REDIS_URI> flushdb
```

dhhyi:
> What would be the best way to trigger that (periodically, manually) in a production environment without connecting via shell?

Hmm, as a project customization, I would run it as an additional script in the nginx container. You have `REDIS_URI` already available as an environment variable and you just have to find a way to install redis-cli.

PR #1515, opened 2023-09-26
chore: add docker image names to compose file

dhhyi:
Enables running the project after checkout fast with `docker compose up` which pulls `latest` images.

[AB#89774](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/89774)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@
 version: '3'
 services:
   pwa:
+    image: intershophub/intershop-pwa-ssr
     build:
       context: .
       args:
@@ -58,6 +59,7 @@ services:
   # </CDN-Example>
 
   nginx:
+    image: intershophub/intershop-pwa-nginx
     build: nginx
     depends_on:
       - pwa


shauke:
What does this mean?
`docker compose up` is pulling the images from dockerhub while `docker compose up --build` is building the images locally from the checked out sources?
What happens if `docker compose up --build` was run (local images exist) and then `docker compose up` is run e.g. with a changed docker compose configuration? What images are used in this case, or better how do I use the local images without rebuilding them? 😉

dhhyi:
> What does this mean?

@shauke This step only means that the local images now have a name that can be evaluated to a remote repository. By default compose images have the name <project>-<service> while project defaults to the folder name. You can check this with `docker compose config --images`

Now a couple of scenarios are valid:
If you don't have local images named the way they are defined and you do `up`, docker compose will search for the images in the remote repository and pull them (that's what I meant in the PR description)

If you `build` or `up --build` it will build the images with the local checkout and tag them.

Unless explicitly requested with `--pull`, docker compose will now always use the local images because they are already available.

shauke (approved):
That started fast. 😁

PR #1531, opened 2023-11-16
refactor: throw error in case recaptcha token is empty

skoch-intershop:
## PR Type

[ ] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

In case that the customer uses an invalid recaptcha sitekey, the recatcha token creation failed and the token is empty. This leads to an incorrect response for all request which are protected by captcha v3.

Issue Number: Closes [91171](https://dev.azure.com/intershop-com/Products/_workitems/edit/91171/)

## What Is the New Behavior?

In case the recaptcha token is empty an TS error is thrown.

[ ] Yes
[ x] No

[AB#91188](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/91188)

### DIFF
--- a/src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
+++ b/src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
@@ -3,8 +3,8 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { FormGroup, Validators } from '@angular/forms';
 import { TranslateModule } from '@ngx-translate/core';
 import { RECAPTCHA_V3_SITE_KEY, ReCaptchaV3Service, RecaptchaV3Module } from 'ng-recaptcha';
-import { timer } from 'rxjs';
-import { filter, switchMap, take } from 'rxjs/operators';
+import { throwError, timer } from 'rxjs';
+import { catchError, filter, switchMap, take, tap } from 'rxjs/operators';
 
 import { DirectivesModule } from 'ish-core/directives.module';
 import { whenTruthy } from 'ish-core/utils/operators';
@@ -25,6 +25,8 @@ import {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CaptchaV3Component implements OnInit {
+  private static readonly RECAPTCHA_ERROR_PREFIX = 'Possible configuration error for applied ReCaptchaV3Service!';
+
   @Input({ required: true }) parentForm: FormGroup;
 
   private destroyRef = inject(DestroyRef);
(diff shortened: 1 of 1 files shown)


src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
```diff
@@ -41,7 +41,18 @@ export class CaptchaV3Component implements OnInit, OnDestroy {
           take(1),
           switchMap(() =>
             timer(0, 2 * (60 - 3) * 1000).pipe(
-              switchMap(() => this.recaptchaV3Service.execute(this.parentForm.get('captchaAction').value))
+              switchMap(() =>
+                this.recaptchaV3Service.execute(this.parentForm.get('captchaAction').value).pipe(
```
SGrueber, line 45: use a helper method to have the code only once

src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
```diff
@@ -41,7 +41,18 @@ export class CaptchaV3Component implements OnInit, OnDestroy {
...
+                  first(),
+                  tap(token => {
+                    if (!token) {
```
SGrueber, line 48: use the rxjs filter operator before the tap,instead of the if 
is it intended to test the result only once?

src/app/core/services/api/api.service.ts
```diff
@@ -80,7 +81,10 @@ export class ApiService {
       // testing token gets 'null' from captcha service, so we accept it as a valid value here
       captchaAction !== undefined
         ? // captcha V3
-          headers.set(ApiService.AUTHORIZATION_HEADER_KEY, `CAPTCHA recaptcha_token=${captcha} action=${captchaAction}`)
+          headers.set(
+            ApiService.RECAPTCHA_TOKEN_HEADER_KEY,
```
SGrueber, line 85: does this changed token header key require a certain ICM version?

SGrueber (changes requested):

src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
skoch-intershop: resolved

src/app/extensions/captcha/shared/captcha-v3/captcha-v3.component.ts
skoch-intershop: It is not necessary to use wether first() as well filter(). Condition will be checked within the tap method.

src/app/core/services/api/api.service.ts
skoch-intershop: The X-Captcha-Token header was introduced with [81572](https://dev.azure.com/intershop-com/Products/_workitems/edit/81572). Fixed versions are 7.10.40.0 and 11.0.12.

SGrueber (changes requested):
pls catch the error in case the sitekey is invalid

PR #1542, opened 2023-11-29
refactor: use server setting for message-to-merchant checkout field

dhhyi:
## PR Type

[x] Feature

## What Is the New Behavior?

In ICM 11 `messageToMerchant` is a flag supplied by the `configurations` call, so the according feature toggle is replaced.

For compatibility with ICM 7.10 an `ICMCompatibilityInterceptor` is provided that can be used to simulate the old behavior where the feature is controlled by a configured feature toggle (see the migration guide for further explanation https://github.com/intershop/intershop-pwa/blob/develop/docs/guides/migrations.md#from-50-to-51).

[x] Yes

## Other Information

requires ICM 11.5 or higher

[AB#93333](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/93333)

### DIFF
--- a/docs/concepts/configuration.md
+++ b/docs/concepts/configuration.md
@@ -165,7 +165,6 @@ Of course, the ICM server must supply appropriate REST resources to leverage fun
 | **B2B Features**             |                                                                                                                                                 |
 | businessCustomerRegistration | Create business customers on registration                                                                                                       |
 | costCenters                  | Cost center feature                                                                                                                             |
-| messageToMerchant            | Write a message to the merchant at checkout                                                                                                     |
 | orderTemplates               | Order template feature                                                                                                                          |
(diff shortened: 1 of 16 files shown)


dhhyi:
@SGrueber @shauke For backwards compatibility I would recommend using an interceptor to projects on ICM 7.10 like this: https://github.com/intershop/intershop-pwa/commit/0bbc991a3af229449ed54c7fc93b8a3551365cd4

SGrueber:
With this PR the feature is no longer working for ICM 7.10 customers. We can either deprecate the feature in the PWA 5.0 and merge it later or provide the suggested interceptor.

shauke (approved):

SGrueber (approved):

PR #1573, opened 2024-01-11
fix: after clicking links the page scrolls up

DilaraGueler:
## PR Type

[X] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[X] Documentation content changes
[ ] Other: 

For accessibility reasons, many links were assigned [routerLink]="[]" to ensure they were clickable. However, this approach often results in an unintended page scroll, causing the page to jump to the top.

## What Is the New Behavior?

To address this issue and improve accessibility, the proposed solution involves replacing the traditional anchor tag with `[routerLink]="[]"` with a more robust alternative. Instead, utilize the `keydown.enter` event in combination with the `tabindex` attribute and the `role` attribute to maintain clickability while avoiding undesired page scrolling.

[ ] Yes
[X] No

## Other Information

[AB#92708](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/92708)


SGrueber (changes requested):
pls add role="button" to the links and mention this also in the documentation

SGrueber (changes requested):
in the cdk tables this solution doesn't seem to work, on enter nothing happens.
I suggest to use buttons here (only in the cdk tables) - and a general styling. 
A first approach could be:  <button
                type="button"
                class="btn btn-link btn-tool"
but the paddings and margins still have to be adapted

SGrueber (changes requested):
Modal handling is still not working properly:
Steps to repeat:
- go to myAccount Addresses
- open a delete confirmation modal using tab and enter
- tab to the cancel button in the modal and click enter
- > the modal closes and opens immediately, so the cancel button doesn't work properly
- > the problem occurs with other modals, too

SGrueber:
closed in favor of #1584

PR #1585, opened 2024-02-05
fix: display of duplicated headers for special logins with activated hydration

shauke:
## PR Type

[x] Bugfix

When loggin in with `/cobrowse` or `/punchout` two headers are rendered for a short time. The simple header of the loading page and the standard header once the user was logged in. After a second the simple header will disappear and everything looks as expected.
This is only visible with Server Side Rendering and the newly introduced hydration of Angular 16 (https://angular.io/guide/hydration).

## What Is the New Behavior?

Switching from the usage of `ngSwitch` to `ngIf` for selecting the different headers solved the rendering problem for the login header combinations, even though the HTML source code does not look better this way. :(

An alternative could have been to disable the hydration for the header.
```
<ish-header ngSkipHydration />
```
But this would also mean to loose these improvements for the header.

[x] No

## Other Information

[AB#93543](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/93543)

### DIFF
--- a/src/app/shell/header/header/header.component.html
+++ b/src/app/shell/header/header/header.component.html
@@ -1,11 +1,12 @@
-<ng-container [ngSwitch]="headerType$ | async">
-  <ish-header-error *ngSwitchCase="'error'" />
-  <ish-header-simple *ngSwitchCase="'simple'" />
-  <ish-header-checkout *ngSwitchCase="'checkout'" />
-
-  <ng-container *ngSwitchDefault>
-    <ish-header-default [isSticky]="isSticky$ | async" [deviceType]="deviceType$ | async" [reset]="reset$ | async" />
-  </ng-container>
+<!-- using ngIf instead of ngSwitch to resolve an issue with hydration and duplicate headers for special logins -->
+<ng-container *ngIf="headerType$ | async as headerType; else defaultHeader">
+  <ish-header-simple *ngIf="headerType === 'simple'" />
+  <ish-header-error *ngIf="headerType === 'error'" />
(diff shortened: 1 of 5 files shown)


src/app/shell/header/header/header.component.html
```diff
@@ -1,11 +1,12 @@
-<ng-container [ngSwitch]="headerType$ | async">
-  <ish-header-error *ngSwitchCase="'error'" />
-  <ish-header-simple *ngSwitchCase="'simple'" />
-  <ish-header-checkout *ngSwitchCase="'checkout'" />
-
-  <ng-container *ngSwitchDefault>
-    <ish-header-default [isSticky]="isSticky$ | async" [deviceType]="deviceType$ | async" [reset]="reset$ | async" />
-  </ng-container>
+<!-- using ngIf instead of ngSwitch to resolve an issue with hydration and duplicate headers for special logins -->
+<ng-container *ngIf="headerType$ | async as headerType; else defaultHeader">
```
SGrueber, line 2: in case the headerType is anything else than simple, error or checkout you don't show the default header now. You could map the headerType to an empty string in the ngOnInit event
 ngOnInit() {
    this.headerType$ = this.appFacade.headerType$.pipe(map type -> ...)

SGrueber (changes requested):

src/app/shell/header/header/header.component.html
shauke: added headerTypes and the check for a valid header type , otherwise default header is used

SGrueber (dismissed):

PR #1603, opened 2024-03-08
perf: improve the performance of 'Static Content Page' component

shauke:
perf: improve the performance of 'Static Content Page' component renderings by optimizing/limiting the '/pagetree' REST calls regarding the used 'depth'

## PR Type

[x] Other: performance improvement

To improve the performance of the 'Static Content Page' component the triggered `/pagetree` REST calls default behavior regarding the used `depth` was changed.
In the past no given 'Navigation Depth' in the ICM backend configuration resulted in no limitation at all ("Define how many levels the navigation tree displays. To show all levels, leave the field empty.").

With the content model adaptions of `icm-as-customization-headless:1.7.0` a depth default value of `5` was introduced and the description was changed accordingly.

## What Is the New Behavior?

In the PWA the rendering was adapted as well so that an empty `NavigationDepth` value of the 'Static Content Page' component now defaults to `0` instead of no depth limitation, that resulted in the whole content page tree being fetched and saved to the state.

[x] Yes

BREAKING CHANGES: An empty 'NavigationDepth' value of the 'Static Content Page' component now defaults to '0' instead of no depth limitation, that resulted in the whole content page tree being fetched and saved to the state (see [Migrations / From 5.0 to 5.1](https://github.com/intershop/intershop-pwa/blob/develop/docs/guides/migrations.md#from-50-to-51) for more details).

To keep the current behavior in an existing project either adapt the `0` default in `pagelet.numberParam('NavigationDepth', 0)"` to a reasonable number or set the 'Navigation Depth' values for all 'Static Content Page' component instances in the ICM backend to reasonable depth values instead of leaving them empty.

## Other Information

Note: Best fitting CMS content model comes with ICM 11 + `icm-as-customization-headless:1.7.0` (and newer versions).


SGrueber (changes requested):
if the defaultValue should always be 0 (if no other value is given) why this isn't handled in the numberParam method?

SGrueber (approved):

PR #1610, opened 2024-03-19
feat: Create Order Template from Order

LucasHengelhaupt:
## PR Type

[x] Feature

The PWA does not yet handle creating an order-template from an existing order.

Issue Number: Closes #

## What Is the New Behavior?

The user can create an order-template from the order details page with all the items from that order.

[x] No

## Other Information

[AB#95134](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/95134)


src/app/extensions/order-templates/shared/order-create-order-template/order-create-order-template.component.html
```diff
@@ -0,0 +1,9 @@
+<button type="submit" name="addProduct" [ngClass]="cssClass" (click)="openModal(modal)">
```
SGrueber, line 1: type="button", remove the name - we don't need it and adapt also the 'basket-create-order-template' in the same way

src/app/extensions/order-templates/shared/order-create-order-template/order-create-order-template.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  templateUrl: './order-create-order-template.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+/**
+ * The Create Order Template from Order displays a button which adds the current order to to a new order template.
+ */
+
+@GenerateLazyComponent()
+export class OrderCreateOrderTemplateComponent {
+  @Input() lineItems: LineItemView[];
```
SGrueber, line 21: required input ?

src/app/extensions/order-templates/shared/order-create-order-template/order-create-order-template.component.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    fixture = TestBed.createComponent(OrderCreateOrderTemplateComponent);
+    component = fixture.componentInstance;
+    element = fixture.nativeElement;
+  });
+
+  it('should be created', () => {
+    expect(component).toBeTruthy();
+    expect(element).toBeTruthy();
+    expect(() => fixture.detectChanges()).not.toThrow();
+  });
```
SGrueber, line 39: add 2 tests
- rendering of the button
- the createOrderTemplatesFromLineItems is triggered if the method createOrderTemplates is called

src/app/extensions/order-templates/store/order-template/order-template.effects.ts
```diff
@@ -90,40 +88,35 @@ export class OrderTemplateEffects {
     )
   );
 
-  addBasketToNewOrderTemplate$ = createEffect(() =>
+  createOrderTemplateFromLineItems$ = createEffect(() =>
```
SGrueber, line 91: pls add tests for this effect - it seems some1 forgot to add them

SGrueber (changes requested):

iwiederhold (changes requested):
Please check the appearance of the button again.

iwiederhold (dismissed):

schadii (approved):

src/assets/i18n/en_US.json
```diff
@@ -240,8 +240,9 @@
   "account.order_template.move.added.text": "{{0}} has been added to <a callback=\"callbackHideDialogModal\" href=\"{{2}}\">{{1}}</a>.",
   "account.order_template.name.error.required": "If you want to create a new order template, please enter a name.",
   "account.order_template.new_from_basket.heading": "Create Order Template from Cart",
-  "account.order_template.new_from_basket_confirm.heading": "The order template \"{{0}}\" has been created and all items from the current cart have been transferred.",
+  "account.order_template.new_from_basket_confirm.heading": "The order template \"{{0}}\" has been created and all items have been transferred.",
   "account.order_template.new_from_order.button.create.label": "Create",
+  "account.order_template.new_from_order.heading": "Create Order Template from Order",
```
mglatter, line 245: Please use sentence case, also in line 247.

M-Behr (approved):

src/assets/i18n/en_US.json
SGrueber: We want keep the application consistent. That's why we should not use the sentence case at one place. Please, let's keep it here as it is and change the notation over the whole application at once.

SGrueber (dismissed):

src/assets/i18n/en_US.json
mglatter: Arvid is already working on the task to unify to sentence case but I doubt that his file already contains these new keys. So I thought we'd better consider this for new keys already. Else it might be overlooked. If it doesn't take effect elsewhere, it's even more an argument for adjusting it while we spot it, isn't it? But you can also check with Arvid. I don't want to provoke any merge conflicts either.

SGrueber (approved):

PR #1650, opened 2024-05-02
perf: disable rendering of view contexts by default

shauke:
* commented out view context integration in HTML templates
* providing an sample import file with the used view contexts for ICM
* added documentation

## PR Type

[x] Other: disabled for better performance

Up till now we render a set of view contexts throughout the PWAs category and product pages. This creates a lot of REST calls and decreases SSR rendering performance in an unnecessary way if this feature is not used.

## What Is the New Behavior?

The view context integration is now commented out and no longer active. It needs to be actively enabled where wanted in a customer project or demo system.

[x] No - if view context rendering was not used

## Other Information

[AB#96291](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/96291)

### DIFF
--- a/src/app/pages/product/product-detail/product-detail.component.spec.ts
+++ b/src/app/pages/product/product-detail/product-detail.component.spec.ts
@@ -5,7 +5,6 @@ import { instance, mock, when } from 'ts-mockito';
 
 import { ProductContextFacade } from 'ish-core/facades/product-context.facade';
 import { findAllCustomElements } from 'ish-core/utils/dev/html-query-utils';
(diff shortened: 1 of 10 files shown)


src/app/pages/product/product-detail/product-detail.component.spec.ts
andreassteinmann, line 1: Does the project need to know (add in documentation) that they should include the view contexts in the tests again?

shauke: I added a line to make aware of that.

PR #1654, opened 2024-05-08
feat: add *ishLazyBrowserView directive

dhhyi:
## PR Type

[X] Feature
[X] Refactoring (no functional changes, no API changes)

## What Is the New Behavior?

Added `*ishBrowserLazyView` directive to lazy load content when it is in the viewport on the browser side. This helps limiting requests made in the SSR process.

[ ] Yes
[X] No

## Other Information

[AB#96475](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/96475)

### DIFF
--- a/src/app/core/directives.module.ts
+++ b/src/app/core/directives.module.ts
@@ -1,5 +1,6 @@
 import { NgModule } from '@angular/core';
 
+import { BrowserLazyViewDirective } from './directives/browser-lazy-view.directive';
 import { ClickOutsideDirective } from './directives/click-outside.directive';
 import { IdentityProviderCapabilityDirective } from './directives/identity-provider-capability.directive';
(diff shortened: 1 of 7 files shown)


shauke:
@dhhyi Just for clarification; the recently viewed component is not an example for "limiting requests made in the SSR process", right? There should not be any recently information on the SSR side that would result in any requests.
It is still a good idea to do the recently required REST requests on the client side only when they come in the viewport though.

dhhyi:
> the recently viewed component is not an example for "limiting requests made in the SSR process", right?

Yes, recently doesn't do any REST requests. On SSR it should not do anything at all as it relates on cookies or session storage. But it's a good example for something that should be lazy loaded on the client as it always is below the fold. The jest test also shows how the directive has to be used in tests to trigger the rendering.

PR #1677, opened 2024-06-06
fix: formly improvements

shauke:
## PR Type

[x] Bugfix
[x] Feature

fix: improved `textarea-description` Formly wrapper and renamed it to `maxlength-description`
* compatibility to be used together with `description` wrapper
* required `maxLength` `props` now
* `props` key to provide an alternative translation key was changed from `customDescription` to `maxLengthDescription`

feat: introduce `labelNoTranslate` and `titleNoTranslate` flag to prevent Formly field label translation
* available for `form-field-horizontal` and `form-field-checkbox-horizontal` Formly wrappers
* removed left over usage of `to.` access in Formly templates, used `props.` instead
* added documentation for new flags

fix: Formly 'ish-select-field' select box does not display updated model value correctly
* the translate extension sets the selected option to the first option instead of the one representing the model value when updating the model dynamically
* works fine on initial load
* reworked the custom extension according to the current Formly example - https://formly.dev/docs/examples/advanced/i18n-alternative/

BREAKING CHANGES: The Formly wrapper `textarea-description` was renamed to `maxlength-description` (see [Migrations / From 5.1 to 5.2](https://github.com/intershop/intershop-pwa/blob/develop/docs/guides/migrations.md#from-51-to-52) for more details).

[x] Yes

## Other Information

[AB#97218](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/97218)


src/app/shared/formly/wrappers/horizontal-wrapper/horizontal-wrapper.component.html
```diff
@@ -1,10 +1,11 @@
 <div class="form-group row" [class.has-error]="showError" [attr.data-testing-id]="keyString + '-wrapper'">
   <label
+    *ngIf="props.label"
     [attr.for]="id"
     class="col-form-label"
     [ngClass]="props.labelClass ? props.labelClass : dprops.labelClass"
-    *ngIf="props.label"
-    >{{ props.label | translate }}<span class="required" *ngIf="props.required && !to.hideRequiredMarker">*</span>
+    >{{ props.labelNoTranslate ? props.label : (props.label | translate)
```
SGrueber, line 7: the labelNoTranslate property should be available in the horizontal-checkbox-wrapper as well and be documented in the formly.md documentation

SGrueber (changes requested):

src/app/shared/formly/wrappers/horizontal-wrapper/horizontal-wrapper.component.html
shauke: Done

SGrueber (dismissed):

PR #1720, opened 2024-10-22
fix: use new redirect url

fgiesske-intershop:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

For a redirect before checkout payment, the customer is redirected to the PSP using the redirect URL that the ICM generates.
When the customer is back in the PWA and the redirect was successful, he can add further products to the basket, which then triggers a new basket validation in the checkout.
This basket validation returns that a new redirect is necessary (depending on the payment connector implementation) but does not return a new redirect URL.

## What Is the New Behavior?

This bug has been fixed in ICM version 12.3.0 and the ICM now returns a new redirect URL on basket validation if adjustments are allowed and if a new redirect is needed.

This PR takes the newly generated URL and stores it in the Store to then perform the redirect with the newly generated URL.

[ ] Yes
[x] No

## Other Information

Requires ICM 12.3.0
Azure DevOps Issue: 97415

[AB#100804](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/100804)

### DIFF
--- a/src/app/core/services/basket/basket.service.ts
+++ b/src/app/core/services/basket/basket.service.ts
@@ -222,7 +222,7 @@ export class BasketService {
    */
   validateBasket(scopes: BasketValidationScopeType[] = ['']): Observable<BasketValidation> {
     const body = {
-      adjustmentsAllowed: !scopes.some(scope => scope === 'All'), // don't allow adjustments for 'All' validation steps, because you cannot show them to the user at once
+      adjustmentsAllowed: true,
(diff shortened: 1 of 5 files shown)


shauke:
A migration note might be good.

shauke:
Closing this PR as it is replaced by #1743.

PR #1731, opened 2025-01-07
feat: save punchout information in session storage instead of in cookies

shauke:
## PR Type

[x] Feature

Customer tested OCI punchout from a punchout system. Login to the webshop was successful but returning to the punchout system from the basket with the HOOK_URL failed.
Reason: The integrated browser within the punchout client is not able to set cookies. But the HOOK_URL is saved as a cookie within the PWA.

## What Is the New Behavior?

To avoid problems with OCI and cXML punchout clients that are not able to set cookies, we now save the punchout information in session storage.

Additionally, errors regarding missing punchout information that is supposed to be available in the session storage (`punchout_SID, punchout_ReturnURL, punchout_HookURL`) are now presented in the UI on basket transfer and in the error log.

[x] No

## Other Information

[AB#102803](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/102803)
[AB#91485](https://dev.azure.com/intershop-com/_workitems/edit/91485)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,11 +14,11 @@ services:
     # command: ['node', 'dist/b2c/run-standalone']
     # command: ['node', 'dist/b2b/run-standalone']
     environment:
-      # ICM_BASE_URL:
+      ICM_BASE_URL: 'https://develop.icm.intershop.de'
       LOGGING: 'true'
       # LOG_ALL: 'false'
       SOURCE_MAPS: 'true'
-      TRUST_ICM: 'true'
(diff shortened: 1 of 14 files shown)


SGrueber (changes requested):
Please adapt the cookie-consent.md file (allowed cookies, PWA required cookies) and the environment.model (allowed cookies) accordingly

SGrueber (approved):

suschneider (approved):

PR #1739, opened 2025-01-16
fix: don't send shipping method in POST items request when ICM runs in single-shipment-mode

rowiish:
## PR Type

[x] Bugfix
[ ] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

The common shipping method is send within the POST /basket/:id/items request every time. 
This is necessary to avoid the creation of multiple buckets if the ICM shipping preference `multipleShipmentsSupported` is set to true. 
However, if the ICM shipping preference `multipleShipmentsSupported` is set to false this causes a 422 response from ICM.

## What Is the New Behavior?

The PWA uses the configuration preference `multipleShipmentsSupported` to distinct between both usecases.
The commonShippingMethod is only send if `multipleShipmentsSupported` is true.

[ ] Yes
[x] No

## Other Information

[AB#103178](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/103178)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -52,7 +52,7 @@ jobs:
         run: npm run build:multi
 
       - name: Upload Build Output
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: dist
           path: dist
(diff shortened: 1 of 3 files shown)


SGrueber (changes requested):
the code looks good, just a small improvement needed

SGrueber (approved):

PR #1743, opened 2025-01-24
fix: fetch updated redirect url before redirecting

SGrueber:
## PR Type

[x] Bugfix

A user who uses a redirect before checkout payment, is redirected to the payment service provider (PSP) using a redirect URL provided by the ICM server during the checkout process.
If the user cancels the checkout process afterwards and adds some product to the basket or changes the basket in any other way the redirect url is not calculated again. In the checkout process he will be redirected to the PSP with the outdated URL. After submitting the order an error occurs because the current basket value differs from the amount that has been authorized by the PSP.

## What Is the New Behavior?

Before the user is redirected to the PSP during the checkout process a REST request is sent to the ICM to fetch the currently valid redirect url.  

[x] No

## Other Information
[AB#97415](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/97415)
[AB#103424](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/103424)


shauke (changes requested):
Looks good as far as I can tell.
But a migration note might be good before merging.

shauke (approved):

PR #1746, opened 2025-01-30
fix: use correct punchout REST API headers for cXML and OCI punchout

shauke:
## PR Type

[x] Bugfix

The method `getCxmlPunchoutSession` used the wrong header for a cXML punchout REST API request.
Also the punchout user methods did not use the correct headers when handling cXML users.

## What Is the New Behavior?

The fitting headers to the according punchout resource are used now.

* introduced header for OCI, cXML and general with according versions
* used punchout resource determination only where necessary

[x] No

## Other Information

[AB#103621](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/103621)

### DIFF
--- a/src/app/extensions/punchout/services/punchout/punchout.service.ts
+++ b/src/app/extensions/punchout/services/punchout/punchout.service.ts
@@ -38,17 +38,36 @@ export class PunchoutService {
   private currentCustomer$ = this.store.pipe(select(getLoggedInCustomer), whenTruthy(), take(1));
 
   /**
-   * http header for Punchout API v2
+   * http header for general Punchout
    */
-  private punchoutHeaderV2 = new HttpHeaders({ Accept: 'application/vnd.intershop.punchout.v2+json' });
+  private punchoutHeader = new HttpHeaders({ Accept: 'application/vnd.intershop.punchout.v2+json' });
 
   /**
-   * http header for cXML Punchout API v3
+   * http header for OCI Punchout
    */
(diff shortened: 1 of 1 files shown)


src/app/extensions/punchout/services/punchout/punchout.service.ts
```diff
@@ -229,10 +248,10 @@ export class PunchoutService {
     return this.currentCustomer$.pipe(
       switchMap(customer =>
         this.apiService.get<PunchoutSession>(
-          `customers/${this.apiService.encodeResourceId(customer.customerNo)}/punchouts/${this.getResourceType(
-            'cxml'
-          )}/sessions/${this.apiService.encodeResourceId(sid)}`,
-          { headers: this.punchoutHeaderV2 }
+          `customers/${this.apiService.encodeResourceId(
+            customer.customerNo
+          )}/punchouts/cxml1.2/sessions/${this.apiService.encodeResourceId(sid)}`,
```
sgrumbachIntershop, line 253: I would probably prefer "`${this.getResourceType('cxml')}`" instead of using the string `cxml1.2` directly.
I thought you was using like "${this.getPunchoutResource(user.punchoutType)}" and therefore said you should use the concrete value.

shauke: We will leave it as it is now after discussion.

PR #1771, opened 2025-03-12
docs: fix dead link detection

LucasHengelhaupt:
## PR Type

[x] Bugfix

Some websites have to be manually excluded from the Docs > "Check Dead Links for Changed Files"-action because the requested website would throw a 403 status code, even if the website is reachable.

## What Is the New Behavior?

Adding a request header with a user-agent to the request makes the request look more like a real browser request, and is therefor less likely to be blocked. This also reduces the number of websites that have to be manually blocked.

The intershop and azure URLs are excluded because they are internal links.
The GitHub URLs are excluded because there are way too many in the files (~3500) and mostly just reference commits, 3rd party licenses and packages.

[x] No

## Other Information

[AB#104907](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/104907)

### DIFF
--- a/docs/check-dead-links.mjs
+++ b/docs/check-dead-links.mjs
@@ -10,11 +10,19 @@ async function mapSeries(iterable, action) {
   }
 }
 
+const axiosConfig = {
+  headers: {
+    'User-Agent':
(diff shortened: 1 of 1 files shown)


shauke (approved):
Nice little improvement. :)

PR #1772, opened 2025-03-14
feat: added `autocomplete` HTML attributes for registration form

LucasHengelhaupt:
## PR Type

[x] Feature

Input fields in the registration form don't have `autocomplete`-HTML tags.

## What Is the New Behavior?

Input fields in the registration form now have `autocomplete`-HTML tags, so they can be more reliable automatically filled.

[x] No

## Other Information

[AB#104968](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/104968)


LucasHengelhaupt:
Because of our configuration-groups a bigger refactoring would be needed to only append the autocomplete-attribute to certain forms. This is necessary because most autocomplete-attributes are for your own personal information like first/last name etc, and you don’t want these attributes on forms that are meant for other users like adding a new user to your company.

PR #1794, opened 2025-04-10
feat: cost center management pagination

chris-santamaria-intershop:
## PR Type

[ ] Bugfix
[x] Feature
[ ] Build-related changes
[ ] CI-related changes
[ ] Other: 

On cost center management page there was just a simple, long list of cost centers that the user could browse through

Issue Number: Closes #103821

## What Is the New Behavior?
On cost center management page, there is now a way to paginate through the list of cost centers. By default there are 25 cost centers per page. Moving from page to page will load the next set of cost centers dynamically if needed.

[ ] Yes
[x] No


M-Behr (changes requested):
It works very well👍

Please make following adjustments:
- use one search field "**Search for ID or name**" because 2 search fields are unusual
- rename line items into **items**

iwiederhold:
Please set the paging to the correct width.
paging

The margins around the search input would also have to be consolidated. (see Order History)
wrapper_cost_center

wrapper_order_history

iwiederhold:
Please add a space or a section or a margin of 30px between H1 div and the search input. Analog to the order history page. 
image

M-Behr (changes requested):
I found a new issue:
I have 17 items. I search for "100" and see 12 items now. When I click on next page or select another page number, it displayes 17 items again. The search term is still visible, but the search in the background seems to be reset.

iwiederhold (dismissed):
👍

M-Behr (approved):
Great! 💯

M-Behr (changes requested):
I found a little issue: If you search for something, which doesn't exists, it displays: "Currently there are no cost centers for your company." This message is displayed when user has no cost centers in general. In the search case, this message is better: "No matching results could be found for your search. Please check the search term." See the behavior in the order history as reference.

That's my last finding. I promise! ;)

shauke (approved):

M-Behr (approved):
Great!

src/assets/i18n/en_US.json
```diff
@@ -1033,7 +1037,7 @@
   "product.image.text.alttext": "product photo",
   "product.instock.text": "In stock",
   "product.itemNumber.label": "Product ID:",
-  "product.items.label": "{{0, plural, one{# list item} other{# list items}}}",
+  "product.items.label": "{{0, plural, one{# item} other{# items}}}",
```
mglatter, line 1066: Value was changed, now translations may need to be adapted as well. Is "item" to be understood in terms of "article" now?

src/assets/i18n/en_US.json
```diff
@@ -141,6 +144,7 @@
   "account.costcenter.manager.label": "Cost center manager",
   "account.costcenter.name.error.required": "Please enter a cost center name.",
   "account.costcenter.name.label": "Name",
+  "account.costcenter.paging.items.label": "{{0, plural, one{# item} other{# items}}}",
```
mglatter, line 148: What does "item" refer to in this context? To pages or paging elements?

mglatter: Please check the translations I provided.

src/assets/i18n/en_US.json
mglatter: Please check the translations I provided.

docs/guides/migrations.md
```diff
@@ -10,17 +10,26 @@ kb_sync_latest_only
...
 The order history listing for the PWA was improved.
-Before, there was only a "load more" button to extend the orders list as long as there were more orders.
-Now, the orders list provides a complete paging bar if there are more then 25 (by default) orders to show.
-This functionality requires ICM 13.1.0 or newer that provides the extended `/orders` REST API that returns the needed paging info.
-PWA projects that do not yet use ICM 13.1.0 or newer will only see the first 25 orders without a hint that there are more relevant orders.
-Searching and filtering works as before but the "load more" button is no longer available.
-For such projects it is probably the best to skip the order history paging commit when migrating to a PWA 8.0.0 until they also migrate to ICM 13.1.0.
+Previously, there was only a "load more" button to extend the order list as long as there were more orders.
+Now, the order list provides a complete paging bar if there are more than 25 orders (by default) to display.
+This functionality requires ICM 13.1.0 or newer that provides the extended `/orders` REST API that returns the needed paging information.
```
mglatter: Why is it not called "Order REST API" as we do in the documentation?

docs/guides/migrations.md
```diff
@@ -10,17 +10,26 @@ kb_sync_latest_only
...
+
+PWA projects that do not use ICM 13.1.0 or newer yet will only see the first 25 orders without a hint that there are additional relevant orders.
+Searching and filtering work as before there, but the "load more" button is no longer available.
+For such projects, it is probably best to skip the order history paging commit when migrating to PWA 8.0.0 until they have also migrated to ICM 13.1.0.
+
+The cost center listing for the PWA was improved.
+It now comes with a complete paging bar if there are more than 25 cost centers (by default) to display.
+In addition, search by cost center ID or name is now supported.
+This functionality requires ICM 13.1.0 or newer that provides the new `/costcenters` REST API version 2.0.0 that supports paging and search.
```
mglatter: Why is it not called "Cost Center REST API" as we do in the documentation?

docs/guides/migrations.md
shauke: I guess since this is a more source code oriented developer documentation I went for mentioning the endpoint that was actually used. If you believe it would be more appropriate to use the official documentation name, we could change it to that. This would be fine with me as well.

docs/guides/migrations.md
shauke: The same as for the Orders REST API question. I am fine with your suggestion here as well.

src/assets/i18n/en_US.json
shauke: In this context it refers to the total of all cost centers.

If I understood IAD right it is supposed to be a general name, so the english "items" would be "Elemente" in german. "Paging-Elemente" sounds to technical or specific in my understanding.

Maybe the screenshot makes it clearer.
grafik

src/assets/i18n/en_US.json
shauke: This was a request from IAD to use "items" as the general naming for all paging totals instead of "list items". So in german the more general  "Elemente" would be the better translation than the more specific "Artikel" I assume.

DDurakISH:
@chris-santamaria-intershop @shauke on basket page the whole cost centers list will be requested still. Are you going to plan this within this PR?

DDurakISH:
@shauke this feature is not complete, if the performance improvement is not implemented on all sides of the PWA.

PR #1798, opened 2025-04-15
fix: hybrid approach improvements

shauke:
## PR Type

[x] Bugfix
[x] Feature
[x] Documentation content changes

The Hybrid Approach did not work as expected with ICM 11+. 

## What Is the New Behavior?

Several changes to make the Hybrid Approach work correctly.

[x] No

### DIFF
--- a/.vscode/intershop.txt
+++ b/.vscode/intershop.txt
@@ -99,6 +99,7 @@ maxlength
 mergeable
 minibasket
 minlength
+mkcert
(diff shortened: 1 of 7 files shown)


dlakaistk:
Thanks!

+1 for this doc to be merged to master. I had to hunt this down, but it would have been faster if this had been merged to develop since ICM 11 has been available for a long time.

PR #1825, opened 2025-06-02
fix: disable luarocks `lua-resty-redis-connector` installation

shauke:
## PR Type

[x] Build-related changes

Currently the NGINX docker image builds fail because of failing downloads from luarocks (see https://github.com/luarocks/luarocks/issues/1797).
Since we are only downloading the `lua-resty-redis-connector` from luarocks this part can be disabled for now since the shared Redis cache integration is not actively used yet in any project.

## What Is the New Behavior?

Not installing the `lua-resty-redis-connector` from luarocks for now to prevent the install problems.

[x] Yes

This change completely disables the possibility to connect to Redis for a shared NGINX pod cache.
Since this feature is not actively used yet in any project, disabling it should not be a problem for now.

## Other Information

[AB#107374](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/107374)

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -7,6 +7,11 @@ kb_sync_latest_only
 
 # Migrations
 
+## From 7.0.0 to 7.1.0
+
+Because of installation problems with the used `luarocks` package manager we disabled the installation of the `lua-resty-redis-connector` that provides the functionality to connect to a shared Redis cache.
+Since this functionality can and was currently not yet used by any project, disabling it is not regarded to as a breaking change.
+
 ## From 6.0.0 to 7.0.0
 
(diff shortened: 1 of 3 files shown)


shauke:
The issue will be resolved with the latest luarocks version 3.12.0 (see https://github.com/luarocks/luarocks/issues/1797#issuecomment-2946606164, https://github.com/luarocks/luarocks/blob/main/CHANGELOG.md#luarocks-3120). 
So we probably need to update the used OpenResty once it is available.

Since the now missing `lua-resty-redis-connector` is not yet actively used we will merge this fix and re-enable it later.

PR #1897, opened 2025-08-21
fix: prevent unnecessary multiple dipatching of suggestions action

skoch-intershop:
### 🚀 Description

After dispatching a search request a suggestions request is unnecessarily executed a second time!

Fixes: #<issue-number>

image

_If this PR contains a breaking change, please describe the impact and migration path for existing applications below._

- [ ] Yes
- [x ] No

---

### Other Information

[AB#109518](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/109518)

### DIFF
--- a/src/app/core/facades/shopping.facade.ts
+++ b/src/app/core/facades/shopping.facade.ts
@@ -257,8 +257,8 @@ export class ShoppingFacade {
 
   suggestResults$(searchTerm: Observable<string>) {
     return searchTerm.pipe(
-      debounceTime(400),
       distinctUntilChanged(),
+      debounceTime(400),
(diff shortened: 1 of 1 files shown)


shauke:
The fix is definitely an improvement and I wanted to approve/merge it.
But I have the feeling the problem needs to be solved differently since logically a suggest request just should not be triggered any more once the user requests a search result.
If for example the user selects an entry from the search suggestions an additional suggestions call is triggered even though it is not wanted/needed any more if the user requests the search result. This is not captured by the current fix.

shauke (approved):

PR #1913, opened 2025-09-19
fix: handle static files with `DEPLOY_URL` path

shauke:
### 🚀 Description

Handle path `DEPLOY_URL` values (e.g. `/shop/`) correct in SSR.
The original intention for the `DEPLOY_URL` were CDN deployments with values like `https://cdn.net/`. The handling of such URLs should not be changed by this fix.

---

### 📝 Notes

* should not influence handling of CDN DEPLOY_URL values

---


SGrueber (approved):

src/ssr/server-scripts/distributor.js
```diff
@@ -62,3 +69,10 @@ function getConcurrencySSR() {
 
   return 2;
 }
+
+// copy from deploy-url.ts
+function getDeployURL() {
```
smoldenhauer-ish: it's a copy - so I suggest to keep the same name getDeployURLFromEnv here.

shauke: This sound like a good idea.

PR #1954, opened 2025-11-11
feat: add pre-commit quality checks

andreassteinmann:
### 🚀 Description

Previously, some quality checks were only executed during CI/CD pipelines after pushing code to GitHub. Developers would:

1. Make changes and push to GitHub
2. Wait for CI/CD to run quality checks
3. If issues were found, fix them locally
4. Push again and wait for CI/CD to re-run

This slow feedback loop was inefficient, often requiring multiple push-fix-push cycles.

---

### 🔍 Detailed Changes

Two important checks were needed:
- **Dead code detection** for all `.ts` files: `npx ts-node scripts/find-dead-code.ts`
- **Testbed cleanup** for `.spec.ts` files: `npm run cleanup-testbed --`

#### First Commit: Initial Approach
Added both checks to the pre-commit hook using `lint-staged`. The `--relative` flag in `.husky/pre-commit` ensured scripts received relative paths (e.g., `src/app/file.ts`) instead of absolute paths, required for the cleanup-testbed script's pattern matching.

#### Second Commit: Optimized Approach
The checks are removed from the pre-commit hook because adding these directly to pre-commit hooks would slow down every commit.

Add `check-lint-staged-extended` command which executes the normal pre-commit hook and both checks in addition on staged files. Used `--concurrent false --relative` flags for proper file path handling. This way, the developer can manually execute the command if needed and the pre-commit hook stays untouched:
`npm run check-lint-staged-extended`

---

### 📝 Notes

@shauke @SGrueber If this is too much for the pre-commit hook, we can reject the PR. 

---

### ✅ Open Tasks

---

### 🔗 Other Information

[AB#111779](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/111779)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "check": "npm-run-all \"lint -- --fix\" format compile build:multi test dead-code update-dockerignore clean-localizations",
     "clean-check": "npm-run-all clean check",
     "check-no-changes": "git diff --exit-code --raw -p --stat",
+    "check-lint-staged-extended": "npx lint-staged && npx lint-staged --concurrent false --relative --config scripts/lint-staged.extended.js",
     "show-ignored-files": "git ls-files . --exclude-standard --others --ignored --directory",
     "changelog": "npm install q --no-save --no-audit --no-fund && npx -p conventional-changelog-cli conventional-changelog -n intershop-changelog.js -i CHANGELOG.md -s",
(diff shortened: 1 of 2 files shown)


SGrueber:
@andreassteinmann If the precommit hook will take longer more developers will switch them off - I'd suggest to run the scripts manually if you have often problems after running CI/CD pipelines 
At least I dislike even currently the duration of the precommit hooks

PR #2004, opened 2026-01-22
ci: workflow update and improvements

shauke:
### 🚀 Description

Updated workflow actions and other improvements

---

### 🔗 Other Information

[AB#113727](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/113727)

[AB#113747](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/113747)


.github/workflows/mark-conflicts.yml
andreassteinmann, line 9: Maybe write it in the same way:
  pull_request:
    types: [opened, synchronize]

shauke: Good hint, I will check which way should be the preferred consistent one.

PR #2057, opened 2026-04-02
feat: localization key for invalid passwords

MatthiasSeidel1976:
### 🚀 Description

Localization key for error code `update_email.error.invalid.password` returned by the improved ICM REST API (ICM > 12.5.0) when "changing" the current e-mail address to the same e-mail address but with a new/different password.

---

### 📝 Notes

If the user updates his email, but uses the same email AND an invalid password, the response will have the new key.

---

[AB#115749](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/115749)

[AB#115876](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/115876)


shauke (changes requested):
Could you please rename the error key, that is returned by ICM for this error, from `password.invalid.password` to `update_email.error.invalid.password`?

This way the localization key would not be removed by the localizations cleanup logic, see https://github.com/intershop/intershop-pwa/blob/develop/scripts/clean-up-localizations.js#L8-L10

It would also be less generic then `password.invalid.password`.

shauke:
@MatthiasSeidel1976 Please let me know once the ICM part is merged and what ICM version will actually include the server side changes. With this information we can then merge this PR.

MatthiasSeidel1976:
On the ICM-Side, the branches are merged! 
The targets where icm-as-master, -support12 and support13.

src/assets/i18n/de_DE.json
```diff
@@ -1544,6 +1540,7 @@
   "suggest.terms.headline": "Suchvorschläge",
   "suggest.view_all.link": "Alle Ergebnisse anzeigen",
   "textarea.max_limit": "Ihre Nachricht: {{0}} Zeichen verbleibend.",
+  "update_email.error.invalid.password": "Bitte geben Sie ein gültiges Kennwort ein.",
```
mglatter, line 1543:
```suggestion
  "update_email.error.invalid.password": "Das eingegebene Kennwort ist nicht korrekt.",
```

src/assets/i18n/en_US.json
```diff
@@ -1544,6 +1540,7 @@
   "suggest.terms.headline": "Search suggestions",
   "suggest.view_all.link": "Show all results",
   "textarea.max_limit": "Your message: {{0}} characters remaining.",
+  "update_email.error.invalid.password": "The password you entered is incorrect.",
```
mglatter, line 1543: For security reasons, is it safe to communicate this? I remember that the teams around Thomas Bergmann tried to avoid error messages that are too telling and would allow for fraud via guessing. Not sure about their context, though.

mglatter (changes requested):

src/assets/i18n/en_US.json
MatthiasSeidel1976: I'm open for suggestions, but the usecase is:

The user is logged in. 
He wants to change his email. The email is also his login. 
The formular requires him to enter his new email and the password. 
The user enters his old email as new email, AND enters a wrong password. 
As result, the message will show up! The user stays logged in!

mglatter: Then it's probably fine to tell the root of the problem.

PR #2072, opened 2026-04-27
feat: hide language switch for single locale configurations

andreassteinmann:
### 🚀 Description

The language switch component is hidden if only one language is available. So it now only renders when multiple locales are available from the ICM server. 

---

### 🔍 Detailed Changes

---

### 📝 Notes

Fix flicker on page load:
- Previously, the component displayed all `fallbackLocales` (configured in environment) at page load before the server configuration loaded. When the server only returned one locale and the code checks to hide the language switch - the language switch was shown for a very short period of time and then it was hidden. 
- Now it waits for `serverConfigurationLoaded$` before rendering to ensure only ICM-provided locales are shown.

---

### ✅ Open Tasks

---

### 🔗 Other Information

[AB#116650](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/116650)

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -141,6 +141,11 @@ Two new adapters have been added - `PaypalGooglePayAdapter` and `PaypalApplePayA
 This functionality requires the **Intershop PPCP Connector version 3.1.0** or higher (ICM 14.2.2+, Google Pay JS API 2.0, Apple Pay JS API 4).
 For details, see the [PayPal Integration Guide](./paypal.md).
 
+**Language switch component**
+
+The behavior of the language switch component was changed.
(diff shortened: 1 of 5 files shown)


shauke:
@SGrueber @andreassteinmann we should discuss the `fallbackLocales` functionality in the context of this PR. This can probably be removed since it did not really work anymore before and not at all with this new functionality.

shauke:
> @SGrueber @andreassteinmann we should discuss the `fallbackLocales` functionality in the context of this PR. This can probably be removed since it did not really work anymore before and not at all with this new functionality.

The `fallbackLocales` are used for other things, e.g. render error pages in certain languages, so we will leave it the way it is.

shauke (approved):

PR #2081, opened 2026-05-06
fix: enable patch open tender after redirect

skoch-intershop:
Redirect before payment method handling was broken. The necessary payment details were no longer sent to the ICM after the redirect. The corresponding effect method has a faulty filtering mechanism.
This issue has been fixed by removing the unnecessary filter.

[AB#116764](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/116764)

### DIFF
--- a/src/app/core/store/customer/basket/basket-payment.effects.ts
+++ b/src/app/core/store/customer/basket/basket-payment.effects.ts
@@ -48,7 +48,7 @@ import {
   updatePaymentInstrumentSuccess,
   updatePaypalCreditCardPaymentInstrument,
 } from './basket.actions';
-import { getCurrentBasket } from './basket.selectors';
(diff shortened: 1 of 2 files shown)


src/app/core/store/customer/basket/basket-payment.effects.ts
```diff
@@ -287,7 +287,6 @@ export class BasketPaymentEffects {
           select(getCurrentBasket),
```
SGrueber: pls revert all changes of this hunk, (select(getCurrentBasketId)) as well as your changes in the test!

SGrueber (changes requested):

SGrueber (approved):

PR #2089, opened 2026-05-18
feat: independed support for Design View and Preview Context

fgiesske-intershop:
### 🚀 Description

Fixes CMS content disappearing in Design View due to SSR hydration mismatch caused by using `DESIGNVIEW` as `PreviewContextID`. Separates Design View mode detection into `DesignViewService`, enabling preview and Design View to work simultaneously.

---

### 🔍 Detailed Changes

- `DesignViewService`: Added `isDesignViewMode` getter/setter with `sessionStorage` persistence and `DesignView` URL parameter detection.
- `PreviewService`: Removed `isDesignViewMode` getter, replaced race/timer approach with synchronous URL parameter parsing, and filters out `DESIGNVIEW` context ID to prevent it from being sent as preview context.
- `ContentDesignViewWrapperComponent`: Switched to `DesignViewService.isDesignViewMode` and added `afterNextRender` to prevent SSR mismatch.
- `PreviewInterceptor`: Removed redundant `!isDesignViewMode` guard since Design View no longer sets a preview context ID.

---

### 📝 Notes

BREAKING CHANGES: 
- Design View integration no longer works with `PreviewContextID=DESIGNVIEW` and was replaced by query param `DesignView`

---

### 🔗 Other Information

[AB#116955](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/116955)
[AB#117174](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/117174)

### DIFF
--- a/src/app/core/utils/design-view/design-view.service.ts
+++ b/src/app/core/utils/design-view/design-view.service.ts
@@ -30,13 +30,33 @@ type ToDVMessageType =
 
 @Injectable({ providedIn: 'root' })
 export class DesignViewService {
+  private designViewMode: boolean;
+
   constructor(
     private router: Router,
     private appRef: ApplicationRef,
     private domService: DomService,
     private store: Store
   ) {
+    if (!SSR && new URLSearchParams(window.location.search).has('DesignView')) {
+      this.setDesignViewMode(true);
+    }
     this.init();
+  }
+
+  private setDesignViewMode(value: boolean) {
+    this.designViewMode = value;
+    if (!SSR) {
+      if (value) {
+        sessionStorage.setItem('DesignViewMode', 'true');
+      } else {
+        sessionStorage.removeItem('DesignViewMode');
+      }
+    }
+  }
+
+  isDesignViewMode(): boolean {
+    return this.designViewMode ?? (!SSR ? sessionStorage.getItem('DesignViewMode') === 'true' : false);
   }
 
   /**
--- a/src/app/core/utils/preview/preview.service.ts
+++ b/src/app/core/utils/preview/preview.service.ts
@@ -1,9 +1,10 @@
 import { ApplicationRef, Injectable } from '@angular/core';
-import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
+import { NavigationEnd, Router } from '@angular/router';
 import { Store, select } from '@ngrx/store';
-import { Subject, delay, filter, first, fromEvent, map, race, switchMap, take, timer, withLatestFrom } from 'rxjs';
+import { Subject, delay, filter, first, fromEvent, map, switchMap, take, withLatestFrom } from 'rxjs';
 
 import { getICMBaseURL } from 'ish-core/store/core/configuration';
+import { DesignViewService } from 'ish-core/utils/design-view/design-view.service';
 import { whenTruthy } from 'ish-core/utils/operators';
 
 interface StorefrontEditingMessage {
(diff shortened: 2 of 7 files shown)


src/app/shared/cms/components/content-design-view-wrapper/content-design-view-wrapper.component.ts
```diff
@@ -4,13 +4,13 @@ import { Observable, combineLatest, map } from 'rxjs';
...
 import { ContentPageletEntryPointView, ContentPageletView } from 'ish-core/models/content-view/content-view.model';
 import { DesignViewService } from 'ish-core/utils/design-view/design-view.service';
-import { PreviewService } from 'ish-core/utils/preview/preview.service';
 
 @Component({
   selector: 'ish-content-design-view-wrapper',
   templateUrl: './content-design-view-wrapper.component.html',
   styleUrls: ['./content-design-view-wrapper.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: { ngSkipHydration: 'true' },
```
SGrueber: ngSkipHydration: 'true' disables hydration for all CMS content in the entire application, even in normal (non-design-view) mode. That's a performance concern — the server-rendered CMS DOM is discarded and re-rendered client-side for every user, not just design view users. You have to find a better solution, e.g. conditionally apply ngSkipHydration only when in design view mode, or to conditionally render this wrapper component only in design view.

src/app/core/utils/design-view/design-view.service.ts
```diff
@@ -30,15 +30,38 @@ type ToDVMessageType =
...
+    if (!SSR) {
+      const urlParams = new URLSearchParams(window.location.search);
+      if (urlParams.has('DesignView')) {
+        this.designViewMode = true;
+      }
+    }
     this.init();
   }
 
+  set designViewMode(value: boolean) {
```
SGrueber: we usually don't use setter and getter methods in services. pls change to setDesignViewMode and isDesignViewMode make them private or if they are public document them

src/app/core/utils/preview/preview.service.ts
```diff
@@ -30,24 +31,16 @@ export class PreviewService {
...
-        timer(3000),
-      ]).subscribe(value => {
-        if (value) {
-          this.previewContextId = value;
-        }
-      });
+      const urlParams = new URLSearchParams(window.location.search);
+      const previewContextId = urlParams.get('PreviewContextID');
+      if (previewContextId && previewContextId !== 'DESIGNVIEW') {
+        this.previewContextId = previewContextId;
```
SGrueber: I actually don't understand this exception. Is it done for backwards compability?
If it is necessary the designView should handle the logic and the preview service should only ask:
if (previewContextId && !this.designViewService.designViewMode) {
  this.previewContextId = previewContextId;
}

src/app/core/utils/preview/preview.service.ts
```diff
@@ -30,24 +31,16 @@ export class PreviewService {
...
-    this.init();
     if (!SSR) {
-      race([
-        this.route.queryParams.pipe(
-          filter(params => params.PreviewContextID),
-          map(params => params.PreviewContextID),
-          take(1)
-        ),
-        // end listening for PreviewContextID if there is no such parameter at initialization
-        timer(3000),
```
SGrueber, line 44: why is this timer no longer needed? It seems to my the previewContextId was not immediately there after pwa start?

SGrueber (changes requested):
please describe your solution in the cms-integration.md and write a short migration note. It seems to me it contains a breaking change, so we can only merge it into the 12.0.0

src/app/core/utils/preview/preview.service.ts
fgiesske-intershop: The new code uses `window.location.search` via `URLSearchParam`, which reads the browser's URL synchronously

src/app/core/utils/preview/preview.service.ts
fgiesske-intershop: I would leave the `PreviewContextID=DESIGNVIEW` in the Design View for now for backwards capability to not break old PWAs.

i can not do `&& !this.designViewService.designViewMode` because it should both work simultaneously

shauke:
Since this is a breaking change anyways and since it is intended for ICM 14+ and IAP Design View we should use this chance to get rid of all the functionality that is still there to work with the ICM 7.10 Design View/Preview/Store Front Editing (SFE).

fgiesske-intershop:
> Since this is a breaking change anyways and since it is intended for ICM 14+ and IAP Design View we should use this chance to get rid of all the functionality that is still there to work with the ICM 7.10 Design View/Preview/Store Front Editing (SFE).

No, the sfe-messages cannot be removed because I have integrated the PWA Storefront Preview into the ICM Backoffice and it uses these messages.

shauke (dismissed):

shauke (dismissed):

mglatter (approved):

shauke (approved):

PR #2101, opened 2026-05-29
chore: Remove identifiable information from nginx and Express responses

kobe-fenego:
### 🚀 Description

Remove identifiable information like version and name from the express and nginx servers.

---

### 🔍 Detailed Changes

Disable the x-powered-by header on the express servers (Angular SSR and theme proxy).
Add server_tokens off; to the nginx.conf so nginx does not return the specific version of openresty.

---

### 📝 Notes

Breaking changes: No
Migrations required: No
Feature flags / configuration changes: None

---

### DIFF
--- a/nginx/nginx.conf.tmpl
+++ b/nginx/nginx.conf.tmpl
@@ -128,6 +128,9 @@ http {
 
     keepalive_timeout 65;
 
+    # Remove the specific version of openresty for security reasons
+    server_tokens off;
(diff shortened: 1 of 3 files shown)


shauke (approved):
@kobe-fenego Thank you for that valuable security contribution.
Once sensitized I added an even more restrictive line before merging it into our source code. :)

PR #2182, opened 2026-09-09
refactor: simplify Store Devtools production configuration

Timerk:
### Description

Prepare for the esbuild migration by configuring Store Devtools through `PRODUCTION_MODE` in `store-devtools.module.ts` instead of an Angular production file replacement in _angular.json_, and the obsolete _store-devtools.module.production.ts_ file has been removed.
The default behavior is unchanged (Store DevTools are enabled in development, disabled in production).

---

### Detailed Changes

- Remove the Store Devtools file replacement from `angular.json` and delete the production-only module.
- Register Store Devtools only when `PRODUCTION_MODE` is false.

### Notes

This change is submitted separately as preparation for the esbuild migration. The webpack build settings remain unchanged.

### Other Information

[AB#120774](https://dev.azure.com/intershop-com/cefd1005-00a7-4c79-927f-a16947d1b2e6/_workitems/edit/120774)


docs/guides/migrations.md
```diff
@@ -9,6 +9,16 @@ kb_sync_latest_only
 
 ## From 12.1.0 to 13.0.0
 
+**Store Devtools production configuration**
```
SGrueber: pls add this migration note at the bottom of this block (from 12.1.0 to 13.0.0), we want to keep most important changes on top

Check and improve the text with langdocks documentation improvement assistant

e2e/cypress/e2e/pages/checkout/cart.page.ts
```diff
@@ -96,16 +96,18 @@ export class CartPage {
   lineItem(idx: number) {
     return {
       quantity: {
-        set: (num: number) =>
-          cy
-            .get(this.tag)
+        set: (num: number) => {
```
SGrueber: this refactoring should not have an effect on e2e tests, pls revert this change

src/app/shared/components/line-item/line-item-list-element/line-item-list-element.component.ts
```diff
@@ -36,8 +37,13 @@ export class LineItemListElementComponent implements OnChanges {
       }
 
       this.updateSubscription = this.context
-        .validDebouncedQuantityUpdate$()
-        .pipe(takeUntilDestroyed(this.destroyRef))
+        .select('quantity')
+        .pipe(
```
SGrueber: this change has nothing to do with the store dev tools, pls revert

SGrueber (changes requested):

e2e/cypress/e2e/pages/checkout/cart.page.ts
Timerk: The change was removed.

src/app/shared/components/line-item/line-item-list-element/line-item-list-element.component.ts
Timerk: The change was removed.

src/app/core/store/store-devtools.module.ts
```diff
@@ -5,13 +5,13 @@ import { loadProductIfNotLoaded } from 'ish-core/store/shopping/products';
...
-  StoreDevtoolsModule.instrument({
-    maxAge: PRODUCTION_MODE ? 25 : 200,
-    logOnly: PRODUCTION_MODE, // restrict extension to log-only mode
-    actionsBlocklist: [loadPromotion.type, loadProductIfNotLoaded.type, setStickyHeader.type, suggestSearch.type],
-    connectInZone: true,
-  }),
-];
+export const storeDevtoolsModule = PRODUCTION_MODE
+  ? []
+  : [
```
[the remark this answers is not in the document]
SGrueber: This is the [official NgRx pattern](https://ngrx.io/guide/store-devtools/recipes/exclude) for excluding devtools in production and has built fine under AOT for a long time.

The Angular compiler runs before PRODUCTION_MODE is substituted (same for webpack and esbuild), so it just re-emits the conditional; the branch is picked at runtime after the bundler inlines the boolean and tree-shakes the unused one. It's also safe in imports because StoreDevtoolsModule adds no declarations, so the compilation scope is unaffected either way.

This lands on the webpack prep branch, so a green production/AOT build in CI confirms it. Happy to switch to a provideStoreDevtools provider later, but it's not needed to fix a bug so I'd keep it out of this migration.

SGrueber (dismissed):

SGrueber (approved):

